Issue #113, opened 2020-10-14
Uploaded camera images are sometimes flipped 90degrees when uploading avatar


adzialocha:

triaslucia:
This is still an issue, but we have no clear reproduction of the issue so far. 
So far we noticed it happens on the first time of uploading a picture, and get corrected when changing it through edit profile flow. It happen on computer.

triaslucia:
I tried to reproduced this bug, on preview with edit profile feature and did not happened again.

triaslucia:
this bug is not possible to be reproduced anymore. Might already been solved, when the "edit profile" got implemented.

PR #122, opened 2020-10-18
fixed: rotate 90deg on avatar placeholder #113

developerfred:
To fix this, it was only necessary to add the rotation in the Avatar placeholder

Screen Shot 2020-10-18 at 11 31 44
Screen Shot 2020-10-18 at 11 31 38

if you were happy with my work you can verify me as a human, i would like to do it at the frontend level i feel bad interacting directly with the contract rs rs

[Hey! Check out my Circles profile!](https://circles.garden/profile/0x8F90cad203C310DaeDC5296bBB26FAcb7920aD17)

### DIFF
--- a/src/components/AvatarUploader.js
+++ b/src/components/AvatarUploader.js
@@ -22,6 +22,7 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: 'transparent',
     border: `1px solid ${theme.palette.text.primary}`,
     cursor: 'pointer',
+    transform: 'rotate(90deg)',
   },
 }));
 


adzialocha:
Thank you @developerfred! I'm afraid though that your suggestion won't work for all cases as the issue is only related to uploaded images on some smartphone devices. With your proposed fix ALL images would be rotated (even the ones who uploaded an image with correct rotation).

I think the fix should actually be applied on the api level (https://github.com/CirclesUBI/circles-api/blob/main/src/middlewares/uploads.js), I can imagine its an issue with reading EXIF meta data and then accordingly handling the image rotation before it gets uploaded to the server.

developerfred:
@adzialocha I'm going to take a look at this api code, thanks for the north.

PR #143, opened 2020-11-27
Move to latest core version

ana0:


adzialocha (changes requested):
Can you run `npm install` as well? Otherwise the version bump is not reflected yet in the `package-lock.json` file :+1:

PR #148, opened 2020-12-15
[Bugfix] Replace styled-components with useStyles

carlfairclough:
This was missing from the recent mnemonic update and is preventing the netlify build

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1647,9 +1647,9 @@
       "dev": true
     },
     "@circles/core": {
-      "version": "2.10.5",
-      "resolved": "https://registry.npmjs.org/@circles/core/-/core-2.10.5.tgz",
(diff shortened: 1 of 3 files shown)


adzialocha:
I'm wondering if we can spare one more dependency to do styles like `styled-components` and use the material-ui system instead? Looks like we would use `styled-components` only for `Mnemonics` component then. I'm happy to introduce it when there is a future path where we want to switch to something else, otherwise I propose to use the material-ui styles API (it even offers a styled-components-like one as well: https://material-ui.com/styles/basics/#styled-components-api).

carlfairclough:
Yeah, will update the PR. I thought styled-components was already in use so didn't see it as a major introduction.

carlfairclough:
PR is updated to drop styled-components and have renamed the PR accordingly

adzialocha (approved):

PR #195, opened 2021-06-21
Retry failed requests when deploying an organization

### LINKED ISSUE #207: Add retry mechanism for failing requests
Due to network connection problems or service outages we face a couple of problems in the app which are not recovered from. We need to a) check for errors b) retry after several attempts to be able to react accordingly.

llunaCreixent:
Sometimes, when onboarding an organization, the process doesn't complete correctly and leaves the user creating a Safe that is neither a UBI-account nor an ORG-account.
We found this problem in several cases when the Safe failed during `organizationSignup` which makes `organization=false`. This PR adds another loop in the organization onboarding process. In the loop, the `signupOrganization` call is reattempted until the the signup is effective in the blockchain. 
This double checking is necessary because sometimes an error is reported: `A server error occurred [connection not open on send()]`. This error occurs when the WebSocket connection to the Ethereum RPC server breaks down, which happens when the Ethereum RPC server is down.
This PR adds a retry on fail function that checks attempts a request until a condition is met. That condition is a function too.
The retry logic is applied to the Safe deployment and the Organization signup.


adzialocha (changes requested):
Cool, thank you for this @llunaCreixent!

I'm not sure if I fully understand what is happening here, from what I can see this is:

1. Calling `request` on every loop iteration (eg. every `LOOP_INTERVAL` seconds). Which means that it tries to call `organization.deploy` every x seconds.
2. It stops doing this when the condition `isOrganization` is true

I think this is sending unnecessary deploy requests EVERY x seconds now to the server, even when no error occurred yet or even when everything is fine and we're just waiting for a result, right? I propose something like this:

1. Call `request` once in the beginning
2. Check frequently for condition, every `LOOP_INTERVAL` seconds
3. When condition is not true after x attempts (like, after 60 seconds etc.), call `request` and start looping again. OR: When an error occurred, catch it and call `request` and start looping again.

A few other small things I've found:

* `utils/helpers` has a little bit of a doubled naming here :smile_cat: maybe pick something more concrete than `helpers`? Like `stateChecks` or `loop` or similar?
* `'~/utils/isDeployed'` is also used in other files (like `src/store/token/actions.js` or `src/store/safe/actions.js`, so the new path needs to be changed there as well)
* `async function loopCheckFunction` we already know that `loopCheckFunction` is a function by its definition, also here I would propose a different name which tells what this is doing. Some ideas: `waitAndRetryOnFail`?

llunaCreixent:
Thanks for the feedback @adzialocha :)
I proposed the name `loopCheckFunction` not to clarify that it's a function but to point out that the loop condition is checked with a function instead of an statement. Yet, I understand it may be confusing.

llunaCreixent:
These changes were done but they must have been dropped in the merge.

>- `'~/utils/isDeployed'` is also used in other files (like `src/store/token/actions.js` or `src/store/safe/actions.js`, so the new path needs to be changed there as well)

llunaCreixent:
I tested locally emulating the network failures using the docker environment (exposing ports) and using `sudo iptables -I DOCKER 1 -p tcp --dport 8545 -j DROP` and `sudo iptables -D DOCKER 1` to loose and recover the connection to the blockchain. 
There is one inconvenience, and is that the steps are not reverted. For example, if the username is already reserved but the Safe is not created, the username could not be used for another account.

adzialocha:
> There is one inconvenience, and is that the steps are not reverted. For example, if the username is already reserved but the Safe is not created, the username could not be used for another account.

Do we need more fine-grained control of retrying? Meaning that we have a retry logic around every step / request, not over the whole collection of requests. This would mean that we need to move the retry mechanism into the core?

llunaCreixent:
> Do we need more fine-grained control of retrying? Meaning that we have a retry logic around every step / request, not over the whole collection of requests. This would mean that we need to move the retry mechanism into the core?

Actually the retry is already covering 1 single step of the onboarding process, so moving the retry to the core won't change anything. The critical point is in the [`createNewOrganization()`](https://github.com/CirclesUBI/circles-myxogastria/blob/main/src/store/onboarding/actions.js#L64) function, where all the steps are carried out sequentially without reverting on failure.

adzialocha:
I see!

> if the username is already reserved but the Safe is not created, the username could not be used for another account.

Maybe good to know: The creation nonce for the safe address is deterministic based on the username, so in that case you would not loose your Safe when trying that step again.

llunaCreixent:
> I see!
> 
> > if the username is already reserved but the Safe is not created, the username could not be used for another account.
> 
> Maybe good to know: The creation nonce for the safe address is deterministic based on the username, so in that case you would not loose your Safe when trying that step again.

But if the onboarding organization process is not finished (for example if the connection to the blockchain is not restored before the retry loop ends) and the user starts the Shared Wallet creation from the first step, the username is lost forever.

adzialocha (approved):
This looks super! Thank you @llunaCreixent!

llunaCreixent:
The Retry is not working properly since the Error is not propagating from the inner loop (of the condition to be checked) to the retry function. Implementing the inner loop directly in the retry function would fix this.
I'm currently working on that.

PR #212, opened 2021-07-06
Consistent rounded `MuiDialog` across all screens

### LINKED ISSUE #211: Update the Issue circles screen’s design
Make the design like presented on the Figma file. 

Wider and one round corner. 

**Design file | Marked with yellow color**

https://www.figma.com/file/xqLKVIJTMAKuwO2wYGbyeb/Shared-Wallets?node-id=1928%3A4

eminx:
This fix changes the global theme configuration of the component `MuiDialog` to the design with the bottom-right corner being rounded at 48px whereas others are straight. 

By this way, without any further changes, all dialogs will be with the same design.


adzialocha (changes requested):
Thanks @eminx! This looks like an elegant way to do this. Unluckily it seems to break some other designs we might need to revisit first:

Screen Shot 2021-07-07 at 15 08 57-fullpage

ps.
I reverted the autolint and package-lock update, as my linter did not warn me about the same things and `npm install` also did not affect the file.

adzialocha:
Ok, the fix was relatively easy, I just moved all action buttons in every Dialog to the left :sparkles:

PR #214, opened 2021-07-07
Retry request logic when failure occurred

llunaCreixent:
The retry function was developed in this PR: https://github.com/CirclesUBI/circles-myxogastria/pull/195
The implementation was incomplete before because the Errors were not propagating in the retry loop. This is fixed now. 

Related issue: https://github.com/CirclesUBI/circles-myxogastria/issues/207 and #216

### DIFF
--- a/src/store/onboarding/actions.js
+++ b/src/store/onboarding/actions.js
@@ -97,9 +97,10 @@ export function createNewOrganization(
           return core.organization.deploy(safeAddress);
         },
         () => {
-          return isOrganization(safeAddress);
+          return core.organization.isOrganization(safeAddress);
         },
       );
+      await isOrganization(safeAddress);
 
       // Prefund the organization with Tokens from the user
       await core.organization.prefund(
--- a/src/store/safe/actions.js
+++ b/src/store/safe/actions.js
@@ -275,9 +275,14 @@ export function deploySafeForOrganization(safeAddress) {
     });
 
     try {
-      await core.safe.deployForOrganization(safeAddress);
-      await isDeployed(safeAddress);
-
+      await waitAndRetryOnFail(
+        () => {
+          return core.safe.deployForOrganization(safeAddress);
+        },
+        async () => {
+          return (await web3.eth.getCode(safeAddress)) !== '0x';
+        },
+      );
       dispatch({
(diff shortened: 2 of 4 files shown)


llunaCreixent:
Also, I changed the implementation of the `loop` to not to use a Promise.

adzialocha:
Thank you @llunaCreixent! I've moved the "nice to have" issue into its own ticket, cleaned up your code a little (removed debugging, simplified logic), added comments and some tests :+1: think this is ready for merging now!

llunaCreixent:
Thanks @adzialocha ! This is great :) 
I was checking that the condition returned true in the `waitAndRetryOnFail` function to also retry if the request wasn't successful. With your implementation I assume we don't take care of what the condition returns because that would be another problem, and not related to network errors, right? I though it's interesting to check for the response being true to proceed with the next steps of the onboarding process, but that can be added in the `actions.js` file.

llunaCreixent:
Also, I wanted to check the condition returning true to prevent the code to continue with the next step without being ready. So this is done with the loop now, right? GREAT!

llunaCreixent:
Super thanks for the tests! 😋

adzialocha:
Yay!

> I was checking that the condition returned true in the waitAndRetryOnFail function to also retry if the request wasn't successful. With your implementation I assume we don't take care of what the condition returns because that would be another problem, and not related to network errors, right? I though it's interesting to check for the response being true to proceed with the next steps of the onboarding process, but that can be added in the actions.js file.

Yes, the API design is a little questionable here, or at least it makes me wanna have TypeScript :rofl:  Currently the retry helper method works best when combined with `loop`, as this one is doing a condition check (not strictly against `true` but something truthy like `1`, a string or true etc.) and throwing an exception when it tried too many times.

I was wondering if we can move the loop directly into the retry helper to not make the user do it themselves, also the API would look similar for both `waitAndRetryOnFail` and `loop`.

I haven't changed it so much yet since I believe that all options are somehow very implicit and hidden, Think even expecting the method to return `true` would be a rather implicit requirement you would only understand by reading the method carefully.. Maybe some more documentation would help and TypeScript one day which cleary leads the developer to use the method the right way.

PR #215, opened 2021-07-07
Organization avatar indicator

### LINKED ISSUE #205: Indicate when profile is an organization
We need some sort of indication if the currently viewed profile is an organization, like a little icon in a corner or something similar.

This should not only be visible on profile pages but also whenever a user profile pops up somewhere (for example when searching for users or when transferring Circles etc.).

This will also help to explain to the users when it is not possible to trust that particular account.

louilinn:
- [x] upload svg
- [x] scale image correctly to the avatars
- [x] add indicator to be responsive to parameter isOrganisation in component (default false)
- [x] everywhere where Avatar component (or AvatarWithQR component) is used send correct isOrganization value to the component


adzialocha:
I've had a look @louilinn and moved all of the `isOrganization` checking straight into the `Avatar` component, even though its different from what we've planned yesterday. Doing it via props feels very verbose and requires loads of code around all other components and views .. think we will have more fun with just doing it inside of `Avatar`.

We can extend the `useIsOrganization` hook to cache its results, so we don't need to make duplicate requests. I've made a ticket here: https://github.com/CirclesUBI/circles-myxogastria/issues/218 don't think it has to be part of this PR as it is an optimization.

louilinn:
I increased the spacing a bit for the avatars on the dashboard view because they got bigger due tot he ring and it looked a bit cramped.

Also worth noting is that the ring does not show in transaction history when two avatars are on top of each other with the org underneath but i think this is fine.

Otherwise it looks good in all the places I could see. So I think I will go ahead and merge this :parrot:

louilinn:
Although the increased spacing makes only 4 avatars fit per row, how do we want to do this @triaslucia 

Does this look ok to you?

Screenshot from 2021-07-08 11-04-06

Before it looked like this:
Screenshot from 2021-07-08 11-10-06

Without increased spacing together with the group wallet indicator it looked like this (but five would fit in a row if there were 5 connections for this user)
Screenshot from 2021-07-08 10-12-04

adzialocha:
@louilinn I think its alright to increase the spacing :+1:

PR #220, opened 2021-07-08
The loading symbol doesn't disappear even after the maximum retry attempts

### LINKED ISSUE #216: The loading symbol doesn't disappear even after the maximum retry attempts
The loading symbol doesn't disappear even after the maximum retry attempts when an Error is caught in the createNewOrganization action in the onboarding store.

Related: #214

llunaCreixent:
With `unlockSafeDeployment()`, the loading view disappears when the `createNewOrganization` organization action fails.
I'm not sure if `checkAppState()` is needed though. 

Closes #216

### DIFF
--- a/src/store/onboarding/actions.js
+++ b/src/store/onboarding/actions.js
@@ -143,8 +143,6 @@ export function createNewOrganization(
     } catch (error) {
       dispatch(hideSpinnerOverlay());
       await dispatch(unlockSafeDeployment());
-      // Force updating app state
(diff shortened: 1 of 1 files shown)


adzialocha (changes requested):
Ah, very nice!

Indeed we don't need `checkAppState` here, it will mostly just check if there are new accounts to connect to, in an error case this might not be required.

adzialocha (approved):

Issue #223, opened 2021-07-09
Inform the user when trying to add an organization as a member to a shared wallet


adzialocha:
This serves as a quick solution for https://github.com/CirclesUBI/circles-myxogastria/issues/192

PR #230, opened 2021-07-15
Change add member dialog

louilinn:
Resolves https://github.com/CirclesUBI/circles-myxogastria/issues/223
There is no longer the option to try and add another shared wallet (previously gave an error)
The shared wallets are still in the search results for adding members but the dialog when clicking them no longer allows you to add them.

Also updated message to inform user adding a member allows them to send circles from shared wallet instead of "issue" which we decided to revert from as I understand it.


adzialocha (changes requested):
Looking good :-)

One small thing: You've created a `sharedWallet.js` hook but I can't see how it is used?

adzialocha (approved):

PR #242, opened 2021-08-09
Update upload error mesage with file requirements

### LINKED ISSUE #202: Clarify image requirements for the avatar
When you upload an image that doesn't have the right format an error shows up saying: `Could not upload image. Is it the correct format and size?`. However the user is not informed about the correct size and format.

llunaCreixent:
Closes https://github.com/CirclesUBI/circles-myxogastria/issues/202

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -35,7 +35,7 @@
     "errorUpdate": "Something went wrong, are you connected to the internet? {errorMessage}"
   },
   "AvatarUploader": {
-    "errorAvatarUpload": "Could not upload image. Is it the right format and size?"
+    "errorAvatarUpload": "Ups, Something went wrong!\nMake sure your image is a jpeg or png of Max 1MB."
   },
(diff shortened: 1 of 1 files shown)


locales/en.json
```diff
@@ -35,7 +35,7 @@
     "errorUpdate": "Something went wrong, are you connected to the internet? {errorMessage}"
   },
   "AvatarUploader": {
-    "errorAvatarUpload": "Could not upload image. Is it the right format and size?"
```
eminx, line 38: Think better with 1 MB rather than 1MB? (with space) :)

eminx (dismissed):

louilinn (dismissed):

locales/en.json
louilinn: I am not sure if "Max" is correct English grammar but maybe "...png with a maximum size of 1 MB"

louilinn (approved):

PR #274, opened 2021-12-17
ci: update actions/setup-node@v2 & simplify cache

Jolg42:
Small change, updating the action and using the native caching, see https://github.com/actions/setup-node#caching-packages-dependencies


louilinn:
Thank you @Jolg42 !

PR #353, opened 2022-05-23
Shared wallet frame size fix

### LINKED ISSUE #327: shared wallet profile frame is too big
There frame on Shared Wallets profile picture are now too big. And they appear floating over other elements. 
Screenshot 2022-04-05 at 14 28 45

### LINKED ISSUE #344: Shared wallet ring indicator sizes are wrong
Shared wallet/organization rings must match the size of the icon it is on

Screenshot from 2022-05-18 12-24-47

mikozet:


src/components/LastInteractions.js
```diff
@@ -87,7 +87,7 @@ const LastInteractionsAccount = ({ address }) => {
   const profilePath = useRelativeProfileLink(address);
 
   return (
-    <Grid item>
+    <Grid item style={{ width: '96px' }}>
```
louilinn, line 90: Can we put this in useStyles instead and give it a clear name / context instead?

louilinn (commented):
Looks like it is fixed! Great :) but I suggest we don't hard code style in the components

src/components/LastInteractions.js
louilinn: I made another PR with a more fundamental fix

louilinn:
I made this PR instead as it fixes the problem more fundamentally instead of fixing it in each component, otherwise this issue will reappear.
https://github.com/CirclesUBI/circles-myxogastria/pull/358

PR #355, opened 2022-05-23
303 list of shared wallets cannot be scrolled

### LINKED ISSUE #303: List of shared wallets cannot be scrolled
**Describe the bug**
In the account switching view https://circles.garden/profile there is no scroll if you have created more shared wallets than can be displayed on the screen. Only solution for the user to reach their bottom accounts is to zoom out which is neither clear nor UX friendly.

**To Reproduce**
1. Go to https://circles.garden/profile
2. If you have many shared wallets (more than 8?) you will notice you cannot scroll to access them all
3. Scrolling works if you have few wallets

I have not reproduced this error but we should try to do it locally (without spamming production with many new SWs)
Martin experienced this bug in mid January 2022 on desktop

mikozet:
Closes: https://github.com/CirclesUBI/circles-myxogastria/issues/303

### DIFF
--- a/src/components/Drawer.js
+++ b/src/components/Drawer.js
@@ -1,4 +1,4 @@
-import { Container, SwipeableDrawer } from '@material-ui/core';
+import { Container, Drawer } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import React from 'react';
 import {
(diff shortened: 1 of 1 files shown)


louilinn (commented):
Have you been able to verify the problem and that it is solved by this change? :) 
I can't reproduce the error atm

mikozet:
Yes I have been able to reproduce it and it is solved now I believe..

śr., 25 maj 2022 o 17:50 louilinn ***@***.***> napisał(a):

louilinn:
I don't want to merge as I cannot verify it is a bug in development. I tried reproducing it both in the browser and on the phone. Can you try again to reproduce on preview.circles.garden?

PR #356, opened 2022-05-24
Adjust BackgroundCurved wave for larger screens

### LINKED ISSUE #347: Make background curve the same height in all views
Screenshot from 2022-05-18 12-39-45
Screenshot from 2022-05-18 12-39-59
 top one (broad view) is too flat, use the same height as in the narrower screen

mikozet:


louilinn:
It still varies a lot :/
Screenshot from 2022-06-09 18-11-41
Screenshot from 2022-06-09 18-11-59

louilinn (approved):
I guess this is good enough for now. 
I would try to have a similar approach as the old tutorial curve if we had more time.

mikozet:
> I would try to have a similar approach as the old tutorial curve if we had more time.

We may try but old tutorial curve is not a curve but circle. That's why in small screens you can see whole circle intead of curve but maybe worth to figure out better solution later on..

PR #357, opened 2022-05-25
349 when scrolling curved background should dissapear and new header bar should appear

### LINKED ISSUE #349: Once scrolling BackgroundCurved background should dissapear and new header bar should appear
@triaslucia could you add turquise version of it in design

mikozet:


louilinn (changes requested):
Looks good when testing in browser! :)
In shared wallet though the color of the personal wallet is universally used instead of the purple.

A question: the bar is shaded horizontally (left to right) while curved background is shaded vertically. Is this intenstional?

PR #366, opened 2022-05-31
Replae overlap=circle by overlap=circular

llunaCreixent:
I get this error in the browser console:

```
Warning: Failed prop type: Material-UI: `overlap="circle"` was deprecated. Use `overlap="circular"` instead.
```


louilinn:
This is fixed in main, that is why i havent fixed in development, butnow it is also fixed in developmen (:

PR #372, opened 2022-06-06
336 change links in menus

### LINKED ISSUE #336: Change links in menus
- in left menu: change Support to "Forum support" (name only changes)
- right buttom menu
  - link to edit page
  - link support to FAQ

mikozet:


louilinn (commented):
See comments in https://github.com/CirclesUBI/circles-myxogastria/pull/374#pullrequestreview-999806623

src/routes.js
```diff
@@ -48,6 +48,7 @@ export const SETTINGS_PATH = '/settings';
 export const SHARE_PATH = '/share';
 export const VALIDATION_PATH = '/validation';
 export const WELCOME_PATH = '/welcome';
+export const EDIT_PROFILE = '/editprofile';
```
louilinn, line 51: When working on the edit profile functionality the route is simply /edit. I think it is better that the route is introduced when we have the functionality

mikozet: Should I maybe remove `edit profile` from menu list?  Or change it for 'edit'

louilinn: It is ok :) I will change it later instead if I want /edit rather than editprofile

PR #374, opened 2022-06-06
337 right bottom menu bottons should be clickable outside text field

### LINKED ISSUE #337: Right bottom menu bottons should be clickable outside text field
only text activates the links

mikozet:


louilinn (commented):
Looking at the code of FloatingNavigation I realize you have built the menu without the existing buttons. What is the reason for that?

I noticed that not the whole span left to right is clickable rather than only the button area itself (which isnt a MUI button).

I think the best solution would be to use the existing button component where all the styling is already done. Atm there is a lot of css styling in the component that I don't see as necessary. We can discuss futher in a call?

mikozet:
> 

Hi @louilinn I changed it to use a Button component, also I removed shadow from list items as in design it was removed..

louilinn (approved):
Great thank you Miko! 
I will merge :)
I think the color on the text seem a bit dark but lets review with Lucia

PR #382, opened 2022-06-10
Fix so that trust button on edit profile does not "Add member"

### LINKED ISSUE #381: Wrong button Add member should be trust
When trusting by pressing on icon on people list - there is Add Member - it should be Trust. It works correctly if you go to user profile and add through icon over there
Zrzut ekranu 2022-06-10 o 11 37 56
Zrzut ekranu 2022-06-10 o 11 37 48
.

louilinn:
current state: extracted add member component but doesnt work in onboarding view

Closes #381

### DIFF
--- a/src/components/Finder.js
+++ b/src/components/Finder.js
@@ -95,7 +95,6 @@ const Finder = ({
   onSelect,
   hasActions,
   filteredSafeAddresses = [],
-  isSharedWalletCreation,
   basePath = SEARCH_PATH,
 }) => {
   const dispatch = useDispatch();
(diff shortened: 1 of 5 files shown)


mikozet:
Hi @louilinn 

I tried to continue from what we have here but didn't make it.
It looks like `MemberSearchAdd` is a wrapper for `Finder` and `DialogAddMember` components which doesnt't bring a lot of optimalisations as there are different functions and props when we use them in `OnboardingOrganization.js` view and `OrganizationMembersAdd.js`

Zrzut ekranu 2022-06-18 o 10 23 07

My proposed solution is here:
https://github.com/CirclesUBI/circles-myxogastria/pull/401

louilinn:
Another solution was merged

PR #386, opened 2022-06-10
Remove zIndex of pink shade

### LINKED ISSUE #385: Pink shade prevents the use of some buttons

louilinn:
Closes #385


mikozet (changes requested):
I added to this few commits.
Basically home screen with links at bottom and I'll do it later text which is clickable on creation organization account

mikozet (approved):

PR #401, opened 2022-06-18
Differentiate Dialog in ProfileMini component depending if we Add mem…

### LINKED ISSUE #381: Wrong button Add member should be trust

mikozet:


src/components/ProfileMini.js
```diff
@@ -133,13 +134,22 @@ const ProfileMini = ({
...
   return (
     <Fragment>
-      <DialogAddMember
-        address={address}
-        handleAddMember={handleAddMember}
-        handleClose={handleClose}
-        isOpen={isOpen}
-        username={username}
-      />
+      {isSharedWalletCreation ? (
```
louilinn, line 137: I think the name of this variable is misleading because it should also chose dialogAddmember if you are in add member view. Thus not only when you create sharedwallet. 
Better name is isModeAddNotTrust or something like that

louilinn (approved):
I think I am confused about why we can't use the same component in the creation view as in the shared wallet add member view. I understand it requires some rework though.
I really like the DRY principle and not having the same code in several places. I think we are creating some technical debt by introduscing the exat same behavior in several components. I think we can do some refactorings later and just make this work for now?

I will merge and we can continue talking about how to improve the code base together later :sunflower:

src/components/ProfileMini.js
mikozet: Maybe `isAddMemberMode`?
We can talk about making code more dry. I don't think we copy same logic several times as we have this `Finder` component which is used in two places. That's why wrapping `Finder` and `DialogAddMember` component as a new component doesn't bring too much value for me. Still there will have to be a logic which replace `DialogAddMember` component as it is use in one place not both. But I may miss something.. 
So discussion is welcomed 🌞

PR #404, opened 2022-06-19
Fix colsole errors in development

louilinn:
Goal is to clean up all console log errors that sometimes appear in local development

### DIFF
--- a/src/components/ActivityIcon.js
+++ b/src/components/ActivityIcon.js
@@ -38,7 +38,12 @@ const DashboardActivityIcon = () => {
       {isPending ? (
         <CircularProgress size={28} />
       ) : (
-        <Badge badgeContent={count} color="primary" max={99}>
+        <Badge
+          badgeContent={count}
+          color="primary"
+          max={99}
+          overlap="rectangular"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -58,8 +58,11 @@
   },
   "dependencies": {
     "@circles/core": "^2.10.10",
+    "@emotion/react": "^11.9.3",
+    "@emotion/styled": "^11.9.3",
     "@material-ui/core": "^4.12.4",
     "@material-ui/lab": "^4.0.0-alpha.61",
+    "@mui/material": "^5.8.4",
```
mikozet: it looks suspicious for me..
so we are adding additional mui library here? Because we still have @material-ui/core...
Here are migrations steps:
https://mui.com/material-ui/migration/migration-v4/

louilinn: I started this migration but it was too much work for now. This dual use fixes the warning though. Do you prefer we fix the warning in another way until we fully migrate mui?

louilinn: I changed to another approach :)

mikozet (approved):

PR #407, opened 2022-06-21
Time circles

### LINKED ISSUE #408: Transition to Time Circles interface
The time circles coversions will happen in the frontend as that is most straight forward based on how things are calculated and the way the code currently looks. The backend deals only with Freckles (10⁻¹⁸ CRC) and Time Circles (TC) are calculated from CRC which are converted in the frontend currently. It is possible that we will refactor this at some point but it is not a priority. Client side conversions have been tested and work smoothly.

louilinn:
- [x] update constants shown in user messages
- [x] display all circles values in time circles
- [x] convert all values to CRC before sending requests to api
- [x] make send amount check in time circles 
- [x] update tests
- [x] add more tests


src/services/activity.js
```diff
@@ -149,7 +149,7 @@ export function formatMessage({
   // Check if find a value in the data (during transfers)
   if ('value' in dataCopy) {
     // Convert the value according to its denominator
-    const valueInCircles = formatCirclesValue(dataCopy.value);
+    const valueInCircles = formatCirclesValue(dataCopy.value, date);
```
llunaCreixent, line 152: Is this the date of the transaction, right? i.e: https://github.com/CirclesUBI/circles-core/blob/main/src/activity.js#L129

src/utils/format.js
```diff
@@ -1,12 +1,12 @@
+import { crcToTc } from '@circles/timecircles';
+
 import web3 from '~/services/web3';
 
-export function formatCirclesValue(value, decimals = 2) {
+export function formatCirclesValue(value, date = Date.now(), decimals = 2) {
```
llunaCreixent: Careful with the order of the parameters. Now that you added `date` as second parameter, anytime you call `formatCirclesValue()` with two params, the second one will be interpreted as `date` and not `decimals`. We must correct all the cases like this: https://github.com/CirclesUBI/circles-myxogastria/blob/time-circles/src/components/UBI.js#L51

src/services/activity.js
llunaCreixent: Have you checked that sometimes we need to use date.now() and sometimes the the when the tx happenes in the past?

src/utils/format.js
```diff
@@ -1,12 +1,12 @@
...
   const valueEth = web3.utils.fromWei(value);
-  const splitted = valueEth.split('.');
-
-  if (splitted.length === 1) {
-    return splitted[0];
-  }
+  // valueEth is a string
+  // crcToTc accepts/converts string to number and returns a number
+  const valueTc = crcToTc(date, valueEth);
```
llunaCreixent: valueEth must be a number, right? 
https://github.com/CirclesUBI/timecircles/blob/master/src/index.ts#L52

src/utils/format.js
```diff
@@ -1,12 +1,12 @@
...
 
-  return `${splitted[0]}.${splitted[1].slice(0, decimals)}`;
+  return valueTc.toFixed(decimals);
```
llunaCreixent: Why we didn't use this function before? 
Maybe bc working with strings is safer?

src/services/activity.js
louilinn: Yes as you can see in the code it is the same date as is displayed as transaction date in the activity log. Transactions dates for the log are provided from the core. 

In all other places I use now. The default is now. And it is only here (for the activity log) that i provide a date. I couldnt find other places that use formatCircleValues ie displays circles that are from the past

louilinn: > Is this the date of the transaction, right? i.e: https://github.com/CirclesUBI/circles-core/blob/main/src/activity.js#L129

Yes!

src/utils/format.js
louilinn: Thank you! Good catch, I looked for these things but somehow missed it. Thank you!

mikozet (approved):

src/views/SendConfirm.js
```diff
@@ -90,13 +91,7 @@ const SendConfirm = () => {
...
 
   const isAmountTooHigh = amount
-    ? web3.utils
-        .toBN(maxAmount)
-        .lte(
-          web3.utils.toBN(
-            amount ? web3.utils.toWei(amount.toString(), 'ether') : '0',
-          ),
-        )
+    ? Number(formatCirclesValue(maxAmount)) < Number(amount)
```
llunaCreixent, line 94: why don't you preserve the comparison in BN?, it should be equivalent as the timestamp is now() for both quatities

llunaCreixent: I think it's safer too compare in Big numbers and not in numbers, so that you don't have to care about decimal inconsistencies. at the end of the day the number in the blockchain is what matter, and the timeCircles conversion functions are consistent (same query, same result)

PR #410, opened 2022-06-22
390 transition screens to new wallet

### LINKED ISSUE #390: Transition screens to New Wallet
We want to support the transition in between Apps for people not to abruptly find a totally new Wallet without advice. 
We will display a set of 3 screens for the first time wallets Login after transition. 

- We use the On-boarding slides system 
- It displays only the first time a Wallet is opens after the transition
-After clicking "Move Forward" people should be re-directed to the landing page of the wallet.
-It is not skip able
-The X icon leads to Login page again. 

Screenshot 2022-06-15 at 12 32 00

**Design Material**  
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=4356%3A49547

### LINKED ISSUE #411: Debug error on development environment
```Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
at BackgroundCurved (http://localhost:8080/app.js:12296:17)
at OnboardingOrganization (http://localhost:8080/app.js:28769:81)
at OnboardingOrganizationContainer (http://localhost:8080/app.js:19829:77)```

mikozet:


louilinn (commented):
Looks great @mikozet !

It is all according to design but I think we should not have the cross (exit) if it shouldnt be skipable. (It is the same thing as skip?)
And I think having back button on slide 2 and 3 makes sense if you wnt to read something again.
The text on slide 2 needs improvement too i think.
I added these comments to Lucia in the issue. Let's wait for the response and change what we agree on

mikozet:
> It is all according to design but I think we should not have the cross (exit) if it shouldnt be skipable. (It is the same thing as skip?)

I believe the same if there is no skip button we should not have x button as well. 
x is the same as skip as it closes tutorial and user will not see it again...

louilinn:
Update from after meeting: https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria
See discussion in the issue.

louilinn (approved):
Perfect! Let's merge :)

PR #428, opened 2022-06-26
Review tc transition

llunaCreixent:
I'm reviewing the TC transition and found some things:
- The following text includes a slash that can be confused with a minus symbol because there are digits in the text too. I find it more clear if we use parenthesis. `you will now receive 1 CRC an hour - 24 a day.`
- Corrected a typo: `bordes` -> `borders`

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -374,7 +374,7 @@
   },
   "TutorialTransition": {
     "slideBody11": "The issuance rate of CRC will increase. Instead of 8 CRC a day, you will now receive 1 CRC an hour (24 CRC a day).",
-    "slideBody12": "With the current changes, Circles users worldwide can share the same value of 1 CRC = 1 hour. This helps to keep the value of UBI more stable across bordes.",
+    "slideBody12": "With the current changes, Circles users worldwide can share the same value of 1 CRC = 1 hour. This helps to keep the value of UBI more stable across borders.",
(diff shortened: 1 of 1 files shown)


llunaCreixent:
Also there is a incoherence in the numbers when:
1. I transfer 10 CRC to someone.
2. Then I see in activity log: "you sent 9.99"

Is this because the rounding is causing the mistake?

mikozet (approved):

mikozet:
> Also there is a incoherence in the numbers when:
> 
> 1. I transfer 10 CRC to someone.
> 2. Then I see in activity log: "you sent 9.99"
> 
> Is this because the rounding is causing the mistake?

I noticed the same ...
It looks like in activity log it always show 0.01 Circles less? as one of my transfers was for 10 Circles and the other for 5.34...

Zrzut ekranu 2022-06-27 o 07 11 25

Is it a potential issue @louilinn , @houndbee ?

louilinn (approved):
Thanks! Removing typos :)

louilinn:
> > Also there is a incoherence in the numbers when:
> > 
> > 1. I transfer 10 CRC to someone.
> > 2. Then I see in activity log: "you sent 9.99"
> > 
> > Is this because the rounding is causing the mistake?
> 
> I noticed the same ... It looks like in activity log it always show 0.01 Circles less? as one of my transfers was for 10 Circles and the other for 5.34...
> 
> Zrzut ekranu 2022-06-27 o 07 11 25
> 
> Is it a potential issue @louilinn , @houndbee ?

Thank you for this! We will create an issue and fix after the release today

PR #432, opened 2022-06-27
Round values in activity log mathematically instead of always rounding down

### LINKED ISSUE #429: Fix rounding decimals in displayed TimeCircles
It looks like in activity log it always show 0.01 Circles less than what I expect to see.
Example:

1. I transfer 10 CRC to someone.
2. Then I see in activity log: "you sent 9.99"

In this image you can see how one of the transfers was for 10 Circles and the other for 5.34:

Zrzut ekranu 2022-06-27 o 07 11 25

louilinn:
Closes #429

### DIFF
--- a/src/services/activity.js
+++ b/src/services/activity.js
@@ -149,7 +149,7 @@ export function formatMessage({
   // Check if find a value in the data (during transfers)
   if ('value' in dataCopy) {
     // Convert the value according to its denominator
-    const valueInCircles = formatCirclesValue(dataCopy.value, date);
+    const valueInCircles = formatCirclesValue(dataCopy.value, date, 2, false);
     dataCopy.denominator = 'Circles';
     dataCopy.value = valueInCircles;
   }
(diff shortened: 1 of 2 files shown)


louilinn:
Before
Screenshot from 2022-06-27 15-19-42
After
Screenshot from 2022-06-27 15-19-34

mikozet:
Do we want to show numbers after decimals?  Currently they are not visible in Activity tab

Zrzut ekranu 2022-06-27 o 15 54 09

louilinn:
> Do we want to show numbers after decimals? Currently they are not visible in Activity tab
> 

When the decimals are .00 js gets no decimals in the string to the roundDown method and by this logic:
```
 if (!decimals) {
    return wholeNumber;
}
```
(from line 8 in format.js)
No decimals are shown for numbers that are exact.
This logic is the same as before. I have not changed it.
See this PR: https://github.com/CirclesUBI/circles-myxogastria/pull/407/files#diff-f1a32ed68bf600a98bd0303b21cc711f7171e3118146f0de2718780ad45bd8ae

mikozet (approved):

llunaCreixent:
But I expect to see 2 decimals in the activity log

PR #438, opened 2022-06-28
Fix decimals in formatCirclesValue

louilinn:
Related #429 

fixed bugs where number of decimals is always 0
added lots of tests

### DIFF
--- a/src/utils/format.js
+++ b/src/utils/format.js
@@ -5,7 +5,7 @@ import web3 from '~/services/web3';
 const roundDownToString = (amount, nbrOfDecimals) => {
   const [wholeNumber, decimals] = amount.toString().split('.');
 
-  if (!decimals) {
+  if (!decimals || decimals == '' || nbrOfDecimals == 0) {
     return wholeNumber;
   }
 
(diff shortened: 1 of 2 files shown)


src/utils/format.js
```diff
@@ -30,5 +31,5 @@ export function formatCirclesValue(
 
   return roundDown
     ? roundDownToString(valueInTimeCircles, nbrOfDecimals)
-    : valueInTimeCircles.toFixed().toString();
+    : valueInTimeCircles.toFixed(nbrOfDecimals).toString();
```
louilinn, line 34: This was the mistake eheh

mikozet:
As general could you link the issue this PR is resolving? Here it is mention in description but I found in other PRs that it is not mentioned

I found an issue when creating shared wallet. Numbers of circles doesn't match.
For example I created shared wallet by sending there 2 circles but at the end created wallet has 5.34
I treid with 7 and 11 and I am getting different numbers after creation:
Zrzut ekranu 2022-06-29 o 13 57 27

louilinn:
@mikozet great find I will resolve this too in another PR

Issue #416, opened 2022-06-23
Remove transition screens after 90 days (25th of September)


louilinn:
Issue for creation https://github.com/CirclesUBI/circles-myxogastria/issues/390

PR creation https://github.com/CirclesUBI/circles-myxogastria/pull/410

PR #440, opened 2022-06-30
415 transitions screens should not show up for new users

### LINKED ISSUE #415: Transitions screens should not show up for new users
Users who registered after the transition date should not see the transition screens

Suggestion check ref link if you came from onboarding page update local storage with transition seen

mikozet:


louilinn (approved):
Code looks good, I dont have time to test the behaviour before vacation but I heard you tested it ;)

Let's remember to remove this as well in this issue #416

PR #447, opened 2022-07-02
Add additional check for fromValidation variable

mikozet:


mikozet:
nobody arround merging

Issue #421, opened 2022-06-23
Validation view: name and avatar not visable until reload


louilinn:
After onboarding, while in validation mode for personal accounts, name and avatar is not visible until page is reloaded

PR #462, opened 2022-07-22
237 edit profile picture and name

mikozet:
Closes https://github.com/CirclesUBI/circles-myxogastria/issues/421


llunaCreixent:
note that the user data cannot be updated until the Safe is deployed and signed up to the Hub

mikozet:
> note that the user data cannot be updated until the Safe is deployed and signed up to the Hub

Thanks for reminding. That case is covered by redirection - if safe is not deployed user will end up on /welcome, /validation pages so I guess we are fine here

llunaCreixent:
I was testing in local dev env, and just after registering (in validation view), The username wasn't there, and instead the safeAddress. Then after I refreshed, I was able to see my username in the validation view.

mikozet:
Hey @llunaCreixent yes we have this registered as an issue here: https://github.com/CirclesUBI/circles-myxogastria/issues/421

locales/en.json
```diff
@@ -109,6 +109,26 @@
...
+  "EditProfile": {
+    "bodyCancel": "If you click cancel all changes you made will be lost.",
+    "buttonCancel": "Cancel",
+    "buttonContinue": "Continue Editing",
+    "buttonSave": "Save changes",
+    "charactersExplanation": "Only basic characters and numbers (A-Z, 0-9) are allowed, no symbols or whitespace, 3-24 characters",
+    "confirmationMessage": "Great! Your profile changes have been saved!",
+    "errorNameTaken": "Sorry, this name is already taken!",
+    "errorInvalidName": "Ups! Invalid name, try again.",
+    "errorImageCapture": "Sorry, there was a problem with image capture",
```
llunaCreixent: Some messages end with period (".") and others not. Shouldn't it be coherent?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "circles-myxogastria",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
llunaCreixent, line 3: If we are adding a new feature the version should be `2.1.0`, right?
Nevertheless, I believe the version is updated after merging to main, isn't it? See https://github.com/CirclesUBI/circles-myxogastria/blob/main/RELEASE.md)

llunaCreixent (commented):
Looks good!! But I added com comments
- ✅ Tests: PASS
- ✅ Lint: didn't pass but I added a commit

package.json
mikozet: I think this is because on main we have version 2.0.1 and development haven't been updated from main so that change is currently visible

llunaCreixent:
If my username was "miko" and I wanted to edit to "Miko", I should be able to do it as long as the username with other case variations is not taken by other account. Right now it says that the username is taken, but actually it's not used by other account, so it should be allowed (unless ther's other user with username "MIKO" or similar)

locales/en.json
mikozet: corrected

llunaCreixent:
> If my username was "miko" and I wanted to edit to "Miko", I should be able to do it as long as the username with other case variations is not taken by other account. Right now it says that the username is taken, but actually it's not used by other account, so it should be allowed (unless ther's other user with username "MIKO" or similar)

That is what I check in the API side: https://github.com/CirclesUBI/circles-api/blob/c68f5b483fd46101394a5fef60e959ad29341659/src/controllers/users.js#L92

But maybe it's ok if it's not in the frontend. MAybe ask @triaslucia ??

llunaCreixent:
Shall we add in the Description of this PR, "Closes https://github.com/CirclesUBI/circles-myxogastria/issues/421" ??

package.json
llunaCreixent: Ok!! So the correct new version will be added once merged to main or development, right?

JacqueGM (approved):
approved now for real, functionality works!

llunaCreixent (approved):

mikozet:
> If my username was "miko" and I wanted to edit to "Miko", I should be able to do it as long as the username with other case variations is not taken by other account. Right now it says that the username is taken, but actually it's not used by other account, so it should be allowed (unless ther's other user with username "MIKO" or similar)

I created issue for that:
https://github.com/CirclesUBI/circles-core/issues/135

mikozet:
> 

no let's leave it I need to test it and probably add a little additional code in one place..
It will come in next PR when I add email ...

PR #463, opened 2022-07-22
457 remove dependency on maxflow calculation when sending transfers

### LINKED ISSUE #457: Remove dependency on MaxFlow calculation when sending transfers
**Is your feature request related to a problem? Please describe.**

Currently, when trying to send a CRC transfer, the user has to wait until the maximum amount has been calculated before the text field to input the value of the CRC transaction is enabled. The text field is not enabled if the maximum amount cannot be calculated successfully(timeout, path too complex, etc.), and the maximum amount shows as 0.

**Describe the solution you'd like**

I suggest removing the dependency and allowing the user send a transaction even if the max amount cannot be calculated successfully.  Also, if the Max amount calculation fails, display a descriptive message such as "Maximum amount cannot be calculated at this time." instead of a "0".

mikozet:


src/views/SendConfirm.js
```diff
@@ -300,9 +308,7 @@ const SendConfirm = () => {
       </View>
       <Footer>
         <Button
-          disabled={
-            !amount || amount <= 0 || isAmountTooHigh || isPaymentNoteInvalid
```
llunaCreixent, line 304: We might want to keep `!amount || amount <= 0`, right?
To be sure that there's an amount to be sent.

houndbee (approved):

louilinn:
This is causing problems because now when there is no path it says the path cannot be calculated. We need to distinguish between 0 value paths and errors in calculating paths.

I will create a new issue to fix this.

Issue #472, opened 2022-08-29
Cache for update user profile doesn't work for SW


llunaCreixent:
I tested uploading an image and changing the username in preview, and  the cache doesn't work for Android at phone, chorme & firefox (from debian).

llunaCreixent:
We found out that it only happens for Shared Wallets

PR #473, opened 2022-08-29
Update how we use cache in DashboardOrganization

mikozet:
Closes https://github.com/CirclesUBI/circles-myxogastria/issues/472


llunaCreixent (commented):
The edition of the profile in SW now works perfect. But when I create a SW, it keeps loading forever, staying in the loading screen.

llunaCreixent (approved):

PR #484, opened 2022-09-23
237 edit profile picture and name

mikozet:


louilinn (approved):
Looks great miko!
Just one question, since you hide the plus icon, when is the plus icon used? Which view?

You can merge to development if plus icon still should be on avatar in some view

mikozet:
> Looks great miko! Just one question, since you hide the plus icon, when is the plus icon used? Which view?
> 
> You can merge to development if plus icon still should be on avatar in some view

So plus button is used in Edit profile page view once you hover on profile picture (+ indicates that there is an action regarding that picture), this behaviour Lucia wanted to remove from Dashboard since you are not editing picture after clicking on it

louilinn:
Thank you miko! Makes total sense!
Also remember to delete the branches when you merge :)

PR #496, opened 2022-10-03
Add user notification aka appnote to login and error view

louilinn:
User notification is not readable if error prevents user from loging in
Before the notifiation was only in dashboard and onboarding


JacqueGM:
all good :)! LGTM 🌻

JacqueGM (approved):

PR #508, opened 2022-10-12
464 refactoring input components new3

### LINKED ISSUE #464: Refactoring Input Components
**Refactoring Input Components**

We should substitute all Input components. 

We should particularly for Error-Input state work on specific changes. 

Design Material: https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=6783%3A10747

mikozet:


louilinn:
I do not see this state, it seems only hoover and active state is used in the app for error message version?

src/components/Input.js
```diff
@@ -4,34 +4,163 @@ import {
...
-
 const useStyles = makeStyles((theme) => ({
   inputLabel: {
+    transform: 'translate(28px, -6px)',
+    background: theme.custom.colors.white,
+    fontSize: '12px',
+    padding: '2px 6px',
+    transition: 'none',
+    color: (isOrganization) =>
+      isOrganization
```
louilinn: these things are repeated a lot. Please define constants for color based on organization and DRY :)

src/components/Input.js
```diff
@@ -4,34 +4,163 @@ import {
...
     },
-  },
-  inputUnderline: {
-    '&.Mui-focused::after': {
-      borderBottomColor: theme.palette.text.primary,
+
+    '&:has(+ .MuiOutlinedInput-root:hover)': {
+      background: (isOrganization) =>
+        isOrganization
+          ? theme.custom.colors.cornflowerBlue
```
louilinn: Also repeated code

louilinn (changes requested):
Great job! Works as intended :) except one comment I already made in this PR. Some small DRY code comments only

mikozet:
>I do not see this state, it seems only hoover and active state is used in the app for error message version?
Uploading Screenshot from 2022-10-14 15-09-15.png…

The link is broken for me but I think I know what do you mean and I apply change - default state for error Input

mikozet:
Hi @louilinn It seems like I didn't push changes before but now they are there ready to review :)

Sorry I still don't see those changes - I am checking now what have happened

mikozet:
> Hi @louilinn It seems like I didn't push changes before but now they are there ready to review :)
> 
> Sorry I still don't see those changes - I am checking now what have happened

Forgot to write - all is good now :)

louilinn:
@mikozet I talked to Rucha and found out that in the latest version on figma, also the search field has this design, so all input fields should have this design so 
- [x] replace search input field
- [x] remove any legacy code connected to the old input field

louilinn:
Also I think something happened.. The hover state in the design is now the active state and the active state is never used.
Current behaviour
writing in field ("active") - has hover design (incorrect)
hovering on field - has hover design (correct)
leaving the field to click on other things and not hovering - has default state (correct)

This is the case for purple (shared wallet) and pink (error) but not blue (private wallet)

louilinn:
I think we can leave the search field update for now but please fix this asap so we can demo:

> Also I think something happened.. The hover state in the design is now the active state and the active state is never used. Current behaviour writing in field ("active") - has hover design (incorrect) hovering on field - has hover design (correct) leaving the field to click on other things and not hovering - has default state (correct)
> 
> This is the case for purple (shared wallet) and pink (error) but not blue (private wallet)

mikozet:
Hi @louilinn it is ready again to review. Bear in mind that input can be active + hovered or just active.

louilinn (commented):
So I'm not sure but feels like if you enter the finder view from shared wallet it should have shared wallet theme?
Or do we always use blue for search?

mikozet:
> So I'm not sure but feels like if you enter the finder view from shared wallet it should have shared wallet theme? Or do we always use blue for search?

You are right it is corrected now

PR #512, opened 2022-10-13
Customizable user notification depending on view in app

louilinn:


src/views/Validation.js
```diff
@@ -146,6 +147,7 @@ const Validation = () => {
       </View>
       <Footer>
         <HumbleAlert>{translate('Validation.bodyDoNotReset')}</HumbleAlert>
```
mikozet: Could we change order here and give AppNote more priority and display it as first alert? Now it is as a second one..

mikozet (approved):

Issue #279, opened 2022-01-05
Names that start with same characters result in same nonce


llunaCreixent:
We found that when the **same owner** creates a SW with either of these names (`test20220104` and `test20220105`), the same address is predicted. It doesn't happen when trying to create the SW from different owner.

We want the nonce of the creation of Shared Wallets (SW) to be deterministic [1], but also different from other nonces used for other safes of the same owner.
We need the nonce to be different to avoid the relayer to find a collision (see [2]).

Possible solution:
- Apply a hashing function to `str` (the name of the SW) in [1] before transform it and splice it.

Topics to be discussed:
- Why do we need the nonce to be deterministic for the SW?

[1] https://github.com/CirclesUBI/circles-myxogastria/blob/3b8a9ba5f8f2c32ddd5304f83a20da90f3750487/src/services/safe.js#L28
[2] https://github.com/CirclesUBI/safe-relay-service/blob/e6844e2b92316ddc099d5b39711487a6e46d5a93/safe_relay_service/relay/views_v3.py#L63

louilinn:
Notes from brainstorming solution: 
The nonce that is created for shared wallets will never be used again once it is deployed (unlike for UBI accounts where you use it to reset from seed phrase)

Making nonce dependent on only name is also a general problem because if you create SW "AAA" change the name to "BBB" you can now not create a new SW called "AAA". This is undesirable. We need to bring in other parameters.

The reason it is deterministic is to make sure it is accessible throughout the deployment process in case browser refreshed etc (this currently breaks but we don't want the nonce to be the reason for it to keep breaking deployment if interupted). Deterministic nonce is not the only way to make the nonce available after a refresh. 

We can make it random (use the already existing createRandomNonce) if we store the nonce in local storage until the SW is deployed. This is ok because the deployment is expected to be finished on the same device and browser where it was started (unlike the UBI account where you await trust connections and deployment can take days)

llunaCreixent:
At least the 9 first characters must be different from any other shared wallet that the user has.
The charcodes are usually 3 characters long, e.g.:
```
"o".charCodeAt() // output: 111
```

llunaCreixent:
Should be closed by https://github.com/CirclesUBI/circles-myxogastria/pull/515

PR #515, opened 2022-10-14
Change how nonce is generated from name

krzysu:
Hi, I looked into https://github.com/CirclesUBI/circles-myxogastria/issues/279 and I found that there is a better way to generate a number (nonce) from a string. I assume that should already solve most of the issues related to the same nonce generated for similar names (strings).
WDYT?

Close https://github.com/CirclesUBI/circles-myxogastria/issues/279

### DIFF
--- a/src/services/safe.js
+++ b/src/services/safe.js
@@ -25,13 +25,15 @@ export function generateDeterministicNonce(address) {
 // Returns a deterministic CREATE2 nonce from a string to predict a
 // to-be-deployed Safe address. This is useful for generating nonces from
 // usernames.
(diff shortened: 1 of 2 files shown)


louilinn:
Thank you @krzysu !
Looks promising! I will await one more review before merging. I will also look more closely at this during the week
I changed the base of this PR to development so we can do proper testing before releasing :)

krzysu:
sure, let me know if there is anything else I could help with!

llunaCreixent (approved):
Thank you very much!

PR #524, opened 2022-10-21
Update .env.example file with example of using banner message

### LINKED ISSUE #523: Update .env.example with using banner content as link

mikozet:


louilinn (commented):
I will merge this to development when I know if we are merging the changes there.

louilinn (approved):

PR #525, opened 2022-10-24
Create mock of data

### LINKED ISSUE #517: Create mockup json data file for news tab
Until api is done create json file for news data
- create a mockup data file to get
- date, text, icon-type, title

[{
id:
title:
text:
url:
date:
icon:
}]

subtask of #396

mikozet:


louilinn:
I had to reset the branches to not include changes in development
If we get merge conflicts (which is hoghly unlikely) we can deal with that later. 
We need to be able to release these changes independently

louilinn (approved):
I would have preferred to put this in a seperate db mock file, and not mix it in with funtionality from the core. I think it would would have been easier to delete. But let's revert this commit later to remove it. Then it will still be smooth
Approved :)

PR #532, opened 2022-11-04
518 expandable news item

### LINKED ISSUE #518: Expandable news item
Subtask of #396 

Make the news items clickable so they exapand to show text and hyperlink / url if included

mikozet:


src/services/core.js
```diff
@@ -255,63 +255,63 @@ const news = {
     {
       id: '1',
       title: 'We have some problems! In two lines, in two lines ',
-      text: 'Lorem ipsum text...',
+      text: '<p>Lorem ipsum text...</p><p>Lorem ipsum text </p>',
```
louilinn, line 258: Shouldn't this be in the component instead? 
I am assuming we don't want the api to respond with html paragraphs?

src/components/NewsFeed.js
```diff
@@ -93,18 +129,53 @@ const NewsList = ({ news }) => {
 
 const NewsItem = ({ newsItem }) => {
```
louilinn, line 130: Shouldn't this be it's own component file?

louilinn (commented):
Some thoughts

src/services/core.js
mikozet: From design perspective in this text section sometimes they include few paragrahs. Probably sometimes they would like to include links if one link at the end is not enough. This can also come from WYSIWYG. Than it is not bad idea to have such a possibility that these is prepared for html content. If it comes as a string we cover that case, if it comes as html we cover that as well...
I wouldn't add it as component as it is typical WYSIWYG the way I see it.

src/components/NewsFeed.js
mikozet: That depends. If you look at it it has 6 lines of code. Do we want to jump through 3-4 components in different files instead of having them in one place? From my perspective those are already prepared to be a separate components (same as NewsItem for example) we can put them in separate files whenever we want but since we are still in development phase of this feature it is a bit easier to navigate when they are in one file..
Will we remember to put them in separate files at the end? Probably not... But also if some of it will grow too much I will add as separate component.

louilinn: I think for me it more has to do with modularity where the code is easy to reuse for another purpose later. But this is maybe fine for now..

src/services/core.js
louilinn: ok, I just thought they would always be just one paragraph and there is a sperate field for urls. 
But yeah this is of course more more future proof for different scenarios.

louilinn (approved):

PR #542, opened 2022-11-23
Add compatibility with Safe 1.3.0

### LINKED ISSUE #541: Enable compatibility with Safe v1.3.0
Tasks:

- [x] Use `circles-core v3.0.0` package
- [x] Migrate the Safe contracts to the latest version, like in this test: https://github.com/CirclesUBI/circles-core/blob/use-latest-safe-contracts/test/safe.test.js#L293

We add a storage variable to hold the Safe version, any time the value is not `1.3.0`, we call the `core.safe.updateSafeVersion` method.
Any time the variable is not in the storage, we call the `core.safe.getVersion` method.

llunaCreixent:
Closes https://github.com/CirclesUBI/circles-myxogastria/issues/541
Closes https://github.com/CirclesUBI/logistics/issues/81


.env.example
```diff
@@ -15,17 +15,21 @@ RELAY_SERVICE_EXTERNAL=http://relay.circles.local
 # Database Endpoints
 DATABASE_SOURCE=graph
 
-# Smart Contract addresses
+# Smart Contract addresses of 1.3.0 version
 HUB_ADDRESS=0xCfEB869F69431e42cdB54A4F4f105C19C080A601
-PROXY_FACTORY_ADDRESS=0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb
-SAFE_ADDRESS=0xC89Ce4735882C9F0f0FE26686c53074E09B0D550
-SAFE_FUNDER_ADDRESS=0x1dF62f291b2E969fB0849d99D9Ce41e2F137006e
```
mikozet, line 22: we don't need SAFE_FUNDER_ADDRESS anymore?

src/services/core.js
```diff
@@ -11,6 +11,7 @@ const core = new CirclesCore(web3, {
   relayServiceEndpoint: process.env.RELAY_SERVICE_EXTERNAL,
   safeMasterAddress: process.env.SAFE_ADDRESS,
   subgraphName: process.env.SUBGRAPH_NAME,
+  fallbackHandlerAddress: process.env.SAFE_DEFAULT_CALLBACK_HANDLER,
```
mikozet: is that being use anywhere?

llunaCreixent: It's being use to initialize the Core

.env.example
llunaCreixent: my bad! thanks for noticing

src/services/core.js
mikozet: aa ok now I got it :) thx

llunaCreixent:
I worrrrks! If you login here: https://devpreview.circles.garden/
You'll see how after a bit you receive a message that your Safe has been updated.

The only thing is that themessage is not ideal haha (users might not know what is a Safe, aaaaannndd... since the message is created before the transaction is approved on blockchain... you can see how the message says version "1.1.1" but then you can check here that the version is 1.3.0 : `https://gnosis-safe.io/app/gno:<SAFE_ADDRESSS>/settings/details`

mikozet:
It updates for me now as you say:
I got message about 1.1.1 version instead of 1.3.0.
Also in my local storage there is new entry - although not with value but promise: safeVersion: [object Promise]
and in my gnosis safe as you say there is version 1.3.0
I am not sure although if we should merge that or tweak a bit more..?

llunaCreixent:
> It updates for me now as you say:
> I got message about 1.1.1 version instead of 1.3.0.
> Also in my local storage there is new entry - although not with value but promise: safeVersion: [object Promise]
> and in my gnosis safe as you say there is version 1.3.0
> I am not sure although if we should merge that or tweak a bit more..?

I think we have to fix the local storage, storing the string and not an object. And the message in the popup... should be more meaningful or even not existing at all... but that is something to decide with @triaslucia .

mikozet (dismissed):

src/components/SafeVersion.js
```diff
@@ -0,0 +1,54 @@
...
+      !safe.currentAccount ||
+      (safe.safeVersion && safe.safeVersion === SAFE_LAST_VERSION)
+    ) {
+      return;
+    }
+
+    const checkSafeVersion = async () => {
+      // Check that the Safe version is the Circles base version
+      if (
+        (await core.safe.getVersion(safe.currentAccount)) !== SAFE_CRC_VERSION
```
llunaCreixent: When is the [`core.safe.getVersion()`](https://github.com/CirclesUBI/circles-core/blob/main/src/safe.js#L42) method executed? Only when I refresh, or anytime I go to any view?
We should think about this; it's important to store the version when it is equal to `1.3.0` so that we avoid making unnecessary eth-calls. But I'm not sure if that is done.
The purpose of the storage or the status was to store the version when `version =1.3.0` in order to avoig the eth calls of calling `core.safe.getVersion()`.
Can you @mikozet confirm that the amount of calls is reduced to the minimum?

llunaCreixent:
It works perfectly! The only thing missing is to confirm that we are saving as much eth_calls as possible, by leveraging the storage and the state.

src/components/SafeVersion.js
mikozet: no, that is not the case we can optimise it more I believe, I will have a look into it

src/store/app/actions.js
```diff
@@ -126,6 +128,8 @@ export function switchAccount(address) {
     }
 
     await dispatch(switchCurrentAccount(address));
+    await removeSafeVersion();
```
louilinn, line 131: why is dispatch not used here?

louilinn (approved):
Tested, works well enough. Some minor fixes will be added in a later release. 

Issue: #546 in particular will be solved later

Another problem I saw was that some accounts don't get the safeVersion field populated in local storage but it flickers by sometimes. Solved by refreshing page. 

Some accounts got 500 response from relayer before the upgrade was complete. Unknown reason.

These things are not so important as it always was successful in the end and this upgrade will only happen once per wallet.

Issue #437, opened 2022-06-28
Refactoring Icons Styles


triaslucia:
We should refactor the style of our icons. Both on Mobile and Desktop version. 

Here Design Material: https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=4356%3A49547

PR #544, opened 2022-11-23
437 refactoring corner icons

louilinn:
Closes: #437 

- [x] replace icons with new ones
- [x] add color scheme
- [x] review white icon logic and simplify if possible/necessary
- [x] (attempt using a gradient on the pink notification badge - cancelled from design)
- [x] remove button area hover effect


triaslucia:
i think we can lose this, as we have new issues for all icons refactoring

louilinn:
This is not an issue but the PR meaning the code related to solve the issue :)
I will change so that PRs are not shown in the zenhub board

mikozet (approved):

PR #557, opened 2022-12-09
Handle no email for user in database in Edit profile view

### LINKED ISSUE #545: Users that don't have an entry in db cannot update profile
**Describe the bug**
The update profile view doesn't load and a completely blank page is displayed.

**To Reproduce**
1. I have an account that doesn't have a user entry in the db (due to a previous issue with the prediction of the safe addresses when validating an account)
2. Click on 'Edit Profile'
3. See error: The edit profile view doesn't load and a completely blank page is displayed. And in Network tab of the browser: **Request**: `https://api.circles.garden/api/users/<safe-address>/email/`, **Response**: `404 Not Found`

**Expected behavior**
Load the page with both "username" and "email" fields empty.

mikozet:
Closes https://github.com/CirclesUBI/circles-myxogastria/issues/545

### DIFF
--- a/src/views/EditProfile.js
+++ b/src/views/EditProfile.js
@@ -375,9 +375,15 @@ const EditProfile = () => {
 
   useEffect(() => {
     (async () => {
-      const userEmail = await core.user.getEmail(safe.currentAccount);
-      setEmailInput(userEmail);
-      setCurrentUserEmail(userEmail);
+      try {
+        const userEmail = await core.user.getEmail(safe.currentAccount);
+        setEmailInput(userEmail);
+        setCurrentUserEmail(userEmail);
+      } catch (error) {
+        logError(error);
+        setEmailInput('');
+        setCurrentUserEmail('');
+      }
     })();
   }, [safe.currentAccount]);
 
--- a/src/views/EditProfile.js
+++ b/src/views/EditProfile.js
@@ -376,14 +376,9 @@ const EditProfile = () => {
   useEffect(() => {
     (async () => {
       if (safe.currentAccount) {
-        const userEmail = await core.user.getEmail(safe.currentAccount);
-        if (userEmail.status !== 404) {
-          setEmailInput(userEmail);
-          setCurrentUserEmail(userEmail);
-        } else {
-          setEmailInput('');
-          setCurrentUserEmail('');
-        }
+        const userEmail = (await core.user.getEmail(safe.currentAccount)) || '';
+        setEmailInput(userEmail);
+        setCurrentUserEmail(userEmail);
       } else {
         setEmailInput('');
         setCurrentUserEmail('');
--- a/src/views/EditProfile.js
+++ b/src/views/EditProfile.js
@@ -375,12 +375,16 @@ const EditProfile = () => {
 
   useEffect(() => {
     (async () => {
-      try {
+      if (safe.currentAccount) {
         const userEmail = await core.user.getEmail(safe.currentAccount);
-        setEmailInput(userEmail);
-        setCurrentUserEmail(userEmail);
-      } catch (error) {
-        logError(error);
+        if (userEmail.status !== 404) {
+          setEmailInput(userEmail);
+          setCurrentUserEmail(userEmail);
+        } else {
+          setEmailInput('');
+          setCurrentUserEmail('');
+        }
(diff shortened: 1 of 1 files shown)


src/views/EditProfile.js
```diff
@@ -375,9 +375,15 @@ const EditProfile = () => {
 
   useEffect(() => {
     (async () => {
-      const userEmail = await core.user.getEmail(safe.currentAccount);
-      setEmailInput(userEmail);
-      setCurrentUserEmail(userEmail);
+      try {
+        const userEmail = await core.user.getEmail(safe.currentAccount);
```
louilinn: Why not only 

`
const userEmail = await core.user.getEmail(safe.currentAccount) || '';`

I don't think we should log this as an error?

louilinn (changes requested):

src/views/EditProfile.js
mikozet: I think catch is more secure here as the line which you have mentioned will not register if it is an error object I believe. 
Generally we should log all type of errors as a good practice as it is easier to debug in future

louilinn: Yes I agree we should log errors but I think for fetching user data that doesnt exist we shouldnt log it as an error. For example when an account doesnt have a usename we don't log it as an error - then the address is displayed instead. Why should a missing email be an error? 
You can see in username.js how we deal with no avatar or no username.
I think we want to make it possible to remove email at some point and also for data erasure requests. These accounts shouldn't then have an email and it is then incorrect to log an error. 
If there is an error with the core call (unlike returning null which isn't an error) there will still be an error? Then we should have try catch around everything in the app which isnt what this issue requests

src/views/EditProfile.js
```diff
@@ -375,9 +375,19 @@ const EditProfile = () => {
 
   useEffect(() => {
     (async () => {
-      const userEmail = await core.user.getEmail(safe.currentAccount);
-      setEmailInput(userEmail);
-      setCurrentUserEmail(userEmail);
+      if (safe.currentAccount) {
+        const userEmail = await core.user.getEmail(safe.currentAccount);
+        if (userEmail.status !== 404) {
```
louilinn: I don't see getEmail ever returning an object with a status attribute?
https://github.com/CirclesUBI/circles-core/blob/e838131b64ae82be2b30fa12e4cb6f2a042f53f0/src/user.js#L270

louilinn (commented):
Would you like to elaborate on why you didn't like the suggestion I proposed?

`const userEmail = await core.user.getEmail(safe.currentAccount) || '';`

mikozet:
> 



> Would you like to elaborate on why you didn't like the suggestion I proposed?
> 
> `const userEmail = await core.user.getEmail(safe.currentAccount) || '';`

I just doesn't feel like it would solve the issue here. If you get 404 from that query how would it save it from crashing?

llunaCreixent (dismissed):
Tested with dockers and correctly reproduced the error with the main branch, and it's solved woth this branch.
I created an account and removed the user entry in the docker db to reproduce the user case.

Good work!

louilinn:
I don't feel that I can comment further on code quality as I cannot test it (without having such an account). So since lluna is the one to test I leave it to her to also review the code quality and if all these if else statements are necessary. And in that case approve the PR @llunaCreixent

llunaCreixent:
> I don't feel that I can comment further on code quality as I cannot test it (without having such an account). So since lluna is the one to test I leave it to her to also review the code quality and if all these if else statements are necessary. And in that case approve the PR @llunaCreixent 

@louilinn you can test locally by removing the entry in the docker database after creating the account.

llunaCreixent:
The solution that @louilinn suggested works better because it doesn't produce an "Uncaught Promise" error.
When the api is down, we get `400` error, When not found we receive `404`. But in any case we display empty email.

llunaCreixent (dismissed):

mikozet:
> The solution that @louilinn suggested works better because it doesn't produce an "Uncaught Promise" error.
> When the api is down, we get `400` error, When not found we receive `404`. But in any case we display empty email.

When we get 404 it is an object so how we will get an empty email? It is a valid response I believe

louilinn (approved):

Issue #436, opened 2022-06-28
Refactoring Buttons Style 


triaslucia:
We should update all main Buttons styles. 

Here design Material: https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=4350%3A18052

mikozet:
from meeting with Lucia: Share Profile double button has the wrong link (it links to send send circles)

PR #559, opened 2022-12-16
Refactor and redesign buttons

mikozet:
closing #436 
https://github.com/CirclesUBI/circles-myxogastria/issues/436


src/styles/theme.js
```diff
@@ -49,6 +49,7 @@ export const colors = {
   silverTree: 'rgba(112, 189, 158, 0.40)',
   finn: 'rgba(90, 47, 86, 0.3)',
   fountainBlue: '#48B2B7',
+  fountainBlue2: '#6CC1C5',
```
louilinn: I don't like this name haha

louilinn: I mean I want it to be descriptive, like is it lighter or what is the difference? is it almost the same? In case it is, should it be the same, is it a design mistake, have you checked?

louilinn (changes requested):
Screenshot from 2022-12-16 15-11-59
Screenshot from 2022-12-16 15-11-35

Button is broken in shared wallet dashboard

src/views/Dashboard.js
```diff
@@ -86,12 +86,25 @@ const useStyles = makeStyles((theme) => ({
   buttonContainer: {
```
louilinn: basically what should be done is to create our own buttonGroup component with the appropriate styling and use that component in all 3 places. That is also part of the refactoring task, to create components for these new styles and not duplicate styling code in all the different places where it is used.

It is also used in MyProfile.js where it is also currently broken

src/styles/theme.js
mikozet: it is lighter, I updated :)

src/views/Dashboard.js
mikozet: I refactor it a bit more

src/components/NavigationFloating.js
```diff
@@ -57,16 +57,16 @@ const useStyles = makeStyles((theme) => {
...
+    const borderFountainBlueLighter = `1px solid ${theme.custom.colors.fountainBlueLighter}`;
     switch (props?.color) {
       case 'turquoise':
-        return borderFountainBlue2;
+        return borderFountainBlueLighter;
       case 'violet':
         return `1px solid ${theme.custom.colors.oldLavender}`;
       case 'white':
-        return borderFountainBlue2;
+        return borderFountainBlueLighter;
```
louilinn: why would case white return the same as turquoise? 
Please just use one, not two cases returning the same

src/components/NavigationFloating.js
```diff
@@ -57,16 +57,16 @@ const useStyles = makeStyles((theme) => {
...
       default:
-        return borderFountainBlue2;
+        return borderFountainBlueLighter;
```
louilinn, line 68: even why have the turquoise case if the default is the same?

src/styles/theme.js
```diff
@@ -30,6 +30,10 @@ const monochrome = {
 const rgba = {
   doveGray: 'rgb(107, 101, 101,0.5)',
   dialogGray: 'rgba(0, 0, 0, 0.25)',
+  silverTree: 'rgba(112, 189, 158, 0.40)',
+  finn: 'rgba(90, 47, 86, 0.3)',
+  greyHover: 'rgba(222, 213, 221, 0.8)',
```
louilinn, line 33: Since greyHover is used in the app and I can't see that view import of this color has changed path it will break

src/styles/theme.js
```diff
@@ -30,6 +30,10 @@ const monochrome = {
...
+  hoverBtn: 'rgba(0, 0, 0, 0.04)',
```
louilinn: I don't think this name "hoverBtn" fits with how we name colors

src/views/Dashboard.js
```diff
@@ -175,14 +145,12 @@ const Dashboard = () => {
...
-              <Button isGradientOpposite isPrimary to={SEND_PATH}>
-                {translate('Dashboard.buttonSendCircles')}
-              </Button>
-            </ButtonGroup>
+            <ButtonTrust
+              leftBtnPath={SEARCH_PATH}
+              leftBtnText={translate('Dashboard.buttonTrustPeople')}
+              rightBtnPath={SEND_PATH}
+              rightBtnText={translate('Dashboard.buttonSendCircles')}
+            />
```
louilinn: :+1:

src/components/ButtonTrust.js
```diff
@@ -0,0 +1,65 @@
...
+
+  buttonGradientOpposite: {
+    background: theme.custom.gradients.purpleOpposite,
+    '&:hover': {
+      background: theme.custom.gradients.purpleOppositeHover,
+    },
+  },
+}));
+
+const ButtonTrust = ({ leftBtnText, rightBtnText }) => {
```
louilinn: This name is not accurate. It is not used for TurstButtons only and even if it was atm it wouldn't be future proof. I suggest ButtonDouble or similar, describing the functionality and not one o its usecases

louilinn (changes requested):
Much better :+1: but I have some comments. I'm sorry to comment on naming but I find it important for code readability.
I will test functionality again once updated

src/styles/theme.js
louilinn: And you can remove silverTree and finn instead of moving them, as they are not used in the app

src/components/NavigationFloating.js
mikozet: For a better code readibility - you know which cases are important

src/styles/theme.js
mikozet: it won't break as you do `...rgba` in `colors` object

src/styles/theme.js
mikozet: corrected

src/styles/theme.js
mikozet: corrected

src/components/NavigationFloating.js
mikozet: corrected

src/components/ButtonTrust.js
mikozet: corrected

louilinn (approved):

PR #562, opened 2022-12-21
Update trust buttons

### LINKED ISSUE #477: Refactoring Trust connections buttons
updating all trust connection buttons. 

**Design Material** https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=7301%3A18288

mikozet:
closes #477 
closes https://github.com/CirclesUBI/circles-myxogastria/issues/477


src/styles/icons.js
```diff
@@ -138,16 +139,20 @@ export const IconTransactions = (props) => {
   return <SvgIcon component={transactions} {...props} viewBox="0 0 20 20" />;
 };
 
+export const IconTrustCustomShape = (props) => {
+  return <SvgIcon component={trustEggShape} {...props} viewBox="0 0 55 55" />;
```
louilinn: I would call it an uneven circle not egg :cat: 
CustomRoundShape
UnevenRoundShape

I would name the icon and svg differently as it is not only for trusting. It is just a shape that is likely to be used elsewhere as we are doing everything hand-drawn long term

src/views/Profile.js
```diff
@@ -426,7 +488,12 @@ const ProfileTrustButton = ({ address, trustStatus }) => {
           {trustStatus.isPending ? (
             <CircularProgress size={24} />
           ) : (
-            <TrustIcon className={classes.trustButtonIcon} />
+            <Box className={classes.trustButtonContainer}>
+              <IconTrustCustomShape
+                className={classes.trustButtonIconBackground}
+              />
+              <TrustIcon className={classes.trustButtonIcon} />
+            </Box>
```
louilinn: I think it would make sense to make this new Uneven Round Button into its own component.

430 lines for one component makes no sense to me. It should be easy to add such a button to another view than Profile.js. Such as the send button in SendConfirm.js for example.

louilinn (changes requested):

src/views/Profile.js
```diff
@@ -470,32 +381,12 @@ const ProfileTrustButton = ({ address, trustStatus }) => {
...
-          ) : (
-            <Box className={classes.trustButtonContainer}>
-              <IconTrustCustomShape
-                className={classes.trustButtonIconBackground}
-              />
-              <TrustIcon className={classes.trustButtonIcon} />
-            </Box>
-          )}
-        </IconButton>
+        <UnevenRoundButton
```
louilinn: Now this UnevenRoundbutton is only for trust.

What I meant was to make a wrapper for IconButton which aways has the uneven round shape inseadof round button. If you look in the designs, I think you will see that this shape is also to be used for send button for example

louilinn (commented):
Thank Miko, but I think you misunderstood the purpose and scope of my proposed UnevenRoundButton component

src/views/Profile.js
louilinn: It is probably a good idea to update send icon button in this PR as well

louilinn (changes requested):
I think you missed my last comments? I don't see any changes or answers.

mikozet:
> I think you missed my last comments? I don't see any changes or answers.

Hey, Could you point me which one do you mean - I think I answered and resolved all of them?
Screenshot from 2023-01-17 11-06-55

louilinn:
Screenshot from 2023-01-18 14-10-57

Nothing changed between the review and you requesting a new one?

mikozet:
> Nothing changed between the review and you requesting a new one?
Sorry, I didn't notice my comment was on pending state...
Screenshot from 2023-01-19 08-22-06

Anyway what do we plan here to do now? Do you want to extend this task for all buttons since now we have confirmation from @ruchajoshi08 about it or do we create a new task?

louilinn:
My suggestion is, as I tried to explain in my comments before as well, to create the UnevenRoundButton component as a generic component not connected to trust buttons in this PR. So that it will be easier to replace the send button and menu button style after. Whether the latter is done in this PR or not is not important to me.

louilinn:
Tasks from product meeting:

4th trust combination possibility missing in design:
- [ ] @triaslucia will provide / confirm design for case: X trust you when you don’t trust them yet. What trust button icon appears on their profile? Make sure it is in the figma file.  (so far we used the pink option but it is not clear in figma)
- [ ] @mikozet update accordingly if necessary

Send button will have this style of wobbly round
- [ ] @mikozet change in this PR
- [ ] @ruchajoshi08 make sure figma file are updated

Three dots menu icon:
- [ ] @triaslucia should it be wobbly round or perfect round?
- [ ] @mikozet update accordingly if necessary

louilinn (approved):

louilinn:
Changing the shape of other round buttons to wobbly will be done in another task. So will reviewing trust button behavior

Issue #580, opened 2023-01-16
Retry on fail for transactions and max flow calculation with fewer max_hops sent to api and pathfinder


louilinn:
We recently upgraded the API (1.8.1) in production to include pathfinder 2.0.
This version allows for a maximum number of trust hops away from them sending user inside the trust network for finding transaction steps. The default number of hops for this version is 3.

The transfer request (to get the transaction steps) often times out for users with complex trust networks. Therefore we want to lower the number of hops and retry finding a path if it times out. Fewer hops requires less computation.
We want to retry with hops = 2 on fail for both max transfer and send.

There are thoughts about saving max hops in local storage as well but not expected in resolving this issue.

louilinn:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Circles-Garden-63764cdd7a20fed2a58bf8cf/issues/CirclesUBI/circles-myxogastria/580?planning-poker) with Zenhub @louilinn @mikozet

llunaCreixent:
What do you think about checking also the steps number to not be more than ~50 (or maybe less, we might have to study the gas limits again) in the new retry method? @louilinn

louilinn:
@llunaCreixent So when we get the error from [here](https://github.com/CirclesUBI/circles-core/blob/cd8c6002c6a4adcf4ace50d303d1b32877e03ee3/src/token.js#L433) we also lower the hops?
I think this will happen automatically as it leads to transfer fail, just like time out?

llunaCreixent:
> @llunaCreixent So when we get the error from [here](https://github.com/CirclesUBI/circles-core/blob/cd8c6002c6a4adcf4ace50d303d1b32877e03ee3/src/token.js#L433) we also lower the hops? I think this will happen automatically as it leads to transfer fail, just like time out?

Sounds good

PR #581, opened 2023-01-20
580 Implement new retry logic for transfers and max flow

### LINKED ISSUE #540: Max amount between shared wallets displaying wrong max flow value
**Describe the bug**

A shared wallet wanting to send CRC to another SW, the max flow amount is displayed incorrectly

**To Reproduce**
1. Go to a shared wallet profile
2. Click on  send circles
3. wait until max amount is displayed 

**Expected behavior**
The max flow amount in time circles

**Screenshots**

Screenshot 2022-11-21 at 18 56 21

**Additional context**
Use this query to check the value returned from the API, this comes in Circles ( so a conversion is needed to time circles)

`{"from":"0x3c89A829400Ea3B49F25738A1F4015A7961D0301","to":"0x9BA1Bcd88E99d6E1E03252A70A63FEa83Bf1208c","value":"1000000000000000000000000000000000"}`

response 
```
{
	"status": "ok",
	"data": {
		"from": "0x3c89A829400Ea3B49F25738A1F4015A7961D0301",
		"to": "0x9BA1Bcd88E99d6E1E03252A70A63FEa83Bf1208c",
		"maxFlowValue": "29100587751936619099862",
		"processDuration": 154335,
		"transferValue": "1000000000000000000000000000000000",
		"transferSteps": []
	}
}
```

louilinn:
- [x] finish core update https://github.com/CirclesUBI/circles-core/pull/153
- [x] add pathfinder hops default constant
- [x] update core service definitions
- [x] new transfer loop
- [x] loop for max flow
- [x] tested on prod data in complex web of trust between shared wallets in preview Netlify deployment
- [x] clean up

Resolves issue: https://github.com/CirclesUBI/circles-myxogastria/issues/580
Closes: https://github.com/CirclesUBI/circles-myxogastria/issues/540#issuecomment-1404436725

### DIFF
--- a/src/store/token/actions.js
+++ b/src/store/token/actions.js
@@ -203,20 +203,68 @@ export function requestUBIPayout(payout) {
   };
 }
 
-async function loopTransfer(from, to, value, paymentNote) {
-  return await waitAndRetryOnFail(
-    () => {
-      return core.token.transfer(from, to, value, paymentNote);
-    },
-    () => {
-      return true;
-    },
-    {},
-    () => {
-      return core.token.updateTransferSteps(from, to, value);
-    },
-  );
+async function transferLoopHops(
+  from,
+  to,
+  value,
+  paymentNote,
+  initialHops = 3,
+  requestedMaxAttempts = 3,
+  waitAfterFail = 5000,
+) {
+  // Count all attempts to retry with fewer hops when something fails
+  let attempt = 1;
+  let hops = initialHops;
+  const maxAttempts = Math.min(initialHops, requestedMaxAttempts);
+  const TRIED_TOO_MANY_TIMES = 'Tried too many times waiting for condition.';
+
+  // Helper method to wait for a few milliseconds before we move on
+  async function wait(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+  }
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    try {
+      // Make transfers request and wait for response
+      const response = core.token.transfer(from, to, value, paymentNote, hops);
+
+      // Return and exit function when the there is no error
+      return response;
+    } catch (error) {
+      // Upon error clean the transfer edges - SHOULD WE DO THIS STILL?
+      // await core.token.updateTransferSteps(from, to, value, hops);
+      if (attempt >= maxAttempts) {
+        // We tried too many times
+        throw error;
+      }
+
+      // Wait when request failed to prevent calling the request too often
+      if (error.message !== TRIED_TOO_MANY_TIMES) {
+        await wait(waitAfterFail);
+      }
+
+      // Lets try again with fewer hops
+      attempt += 1;
+      hops -= 1;
+    }
+  }
 }
+
+// async function loopTransfer(from, to, value, paymentNote) {
+//   return await waitAndRetryOnFail(
+//     () => {
+//       return core.token.transfer(from, to, value, paymentNote);
+//     },
+//     () => {
+//       return true;
+//     },
+//     {},
+//     () => {
+//       return core.token.updateTransferSteps(from, to, value);
+//     },
+//   );
+// }
 
 /**
  * Transfer circles to another safe
@@ -225,7 +273,7 @@ async function loopTransfer(from, to, value, paymentNote) {
  * @param {string} paymentNote Message for recipient
  * @returns response
  */
-export function transfer(to, amount, paymentNote = '') {
+export function transfer(to, amount, paymentNote = '', hops = 3) {
   return async (dispatch, getState) => {
     dispatch({
       type: ActionTypes.TOKEN_TRANSFER,
@@ -238,7 +286,7 @@ export function transfer(to, amount, paymentNote = '') {
       const value = new web3.utils.BN(
         core.utils.toFreckles(tcToCrc(Date.now(), Number(amount))),
       );
-      const txHash = await loopTransfer(from, to, value, paymentNote);
+      const txHash = await transferLoopHops(from, to, value, paymentNote, hops);
 
       dispatch(
         addPendingActivity({
--- a/src/store/token/actions.js
+++ b/src/store/token/actions.js
@@ -313,7 +313,9 @@ async function loopTransfer(
       paymentNote,
       hops,
     });
-    return core.token.transfer(from, to, value, paymentNote, hops);
+    const response = core.token.transfer(from, to, value, paymentNote, hops);
+    console.log('Print after core command within "try"', { response });
+    throw new TransferError();
   } catch (error) {
     console.log('ERROR: ', error);
     // if api times out or there are too many steps to fit in one transfer
--- a/src/store/token/actions.js
+++ b/src/store/token/actions.js
@@ -215,7 +215,7 @@ async function loopTransfer(
   errorsMessages = '',
 ) {
   if (attemptsLeft === 0 || hops === 0) {
-    // run out of attempts or hops, cannot attempt further transfers
+    // ran out of attempts or hops, cannot attempt further transfers
     throw new TransferError(errorsMessages);
   }
   try {
@@ -250,7 +250,7 @@ async function loopTransfer(
     ) {
       // update the edges db for trust-adjacent safes
       await core.token.updateTransferSteps(from, to, value, hops);
-      // try only once after update with the same params
+      // try again after update with the same parameters
       return await loopTransfer(
         from,
         to,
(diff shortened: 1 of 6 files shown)


src/store/token/actions.js
```diff
@@ -203,29 +208,168 @@ export function requestUBIPayout(payout) {
...
+  try {
+    console.log('Attempting with params ', {
+      from,
+      to,
+      value,
+      paymentNote,
+      hops,
+    });
+    return core.token.transfer(from, to, value, paymentNote, hops);
+  } catch (error) {
```
louilinn: I don't seem to be able to catch the 504

llunaCreixent: try now after changing the timeout in the api

louilinn:
Example of successful max flow calculation is shared internally (sensitive business data). It was successful on second attempt with 2 hops between 2 shared wallets with complex connections. Max value is very big and expected to be of this approximate size.

src/utils/findPath.js
```diff
@@ -1,5 +1,52 @@
 import core from '~/services/core';
 import web3 from '~/services/web3';
+import { PATHFINDER_HOPS_DEFAULT } from '~/utils/constants';
+
+const { ErrorCodes } = core.errors;
+const LARGE_AMOUNT = new web3.utils.BN(
+  web3.utils.toWei('1000000000000000', 'ether'),
```
mikozet, line 7: why particularly this value 1000000000000000? maybe we could add comment?

mikozet, line 11: why particularly this value 1000000000000000? maybe we could add comment?

src/store/token/actions.js
```diff
@@ -203,29 +204,86 @@ export function requestUBIPayout(payout) {
...
+  to,
+  value,
+  paymentNote,
+  hops,
+  attemptsLeft,
+  errorsMessages = '',
+) {
+  if (attemptsLeft === 0 || hops === 0) {
+    // ran out of attempts or hops, cannot attempt further transfers
+    throw new TransferError(errorsMessages);
```
mikozet, line 219: should we here display message for users?

src/store/token/actions.js
mikozet: what was the reason console.log hasn't been displayed? are they now?

src/store/token/actions.js
```diff
@@ -203,29 +204,86 @@ export function requestUBIPayout(payout) {
...
+        errorsMessages.concat(' ', error.message),
+      );
+    }
+    // if the path is found but it is invalid
+    else if (
+      (error.name === 'TransferError' && ErrorCodes.TRANSFER_NOT_FOUND) || // search complete and no path found
+      error.name !== 'TransferError' // includes errors from attempting transfer with an invalid path
+    ) {
+      // update the edges db for trust-adjacent safes
+      await core.token.updateTransferSteps(from, to, value, hops);
```
mikozet: shouldn't we have try catch here? Otherwise this query goes silent once there is error from it?

src/store/token/actions.js
```diff
@@ -203,29 +204,86 @@ export function requestUBIPayout(payout) {
...
+    // ---
+    // no path or path too long
+    if (
+      error.name === 'TransferError' &&
+      (error.code === ErrorCodes.TOO_COMPLEX_TRANSFER || // too many steps in found path
+        error.code === ErrorCodes.UNKNOWN_ERROR || // includes timeout error from api
+        error.code === ErrorCodes.INVALID_TRANSFER) // other errors from find transitive transfer
+    ) {
+      // retry with fewer hops
+      return await loopTransfer(
```
mikozet, line 236: maybe would be worth to update the edges of db here as well before looping again?

src/store/token/actions.js
```diff
@@ -203,29 +204,86 @@ export function requestUBIPayout(payout) {
...
+        value,
+        paymentNote,
+        hops,
+        attemptsLeft - 1,
+        errorsMessages.concat(' ', error.message),
+      );
+    }
+    // any other errors will result in propagating the error
+    else {
+      throw error;
```
mikozet: what about information for users? shouldn't we display a message for them?

src/store/token/actions.js
louilinn: Yes it changed to 422 but that is caught in core and core then throws new/other errors according to here: https://github.com/CirclesUBI/circles-core/blob/b661de206bbaaa3a8aca4cc289e0539c48a8600e/src/token.js#L562

src/store/token/actions.js
louilinn: This error is caught (i think in SendConfirm). I kept these as a debug log as the error will go in the logs. I will also add some kind of interpretation to the user here https://github.com/CirclesUBI/circles-myxogastria/blob/29599c9ae2e28ac7e1c0aaa961f641cc858a769e/src/views/SendConfirm.js#L137 Thanks!

src/store/token/actions.js
louilinn: Yes I thought about it.. but wasn't sure how to do it in a good why, what to do if it fails?

src/store/token/actions.js
louilinn: no because it will take more time, and there is no indication that edges are outdated, just that it is either too many hops or takes to long to calculate. most likely update edges will also time out due to the same reason. We use update when a found path is not accepted as transfer

src/utils/findPath.js
louilinn: I thought large amount as name was enough but I can add comment :)

src/store/token/actions.js
louilinn: This error is caught (i think in SendConfirm). I kept these as a debug log as the error will go in the logs. I will also add some kind of interpretation to the user here https://github.com/CirclesUBI/circles-myxogastria/blob/29599c9ae2e28ac7e1c0aaa961f641cc858a769e/src/views/SendConfirm.js#L137 Thanks!

src/store/token/actions.js
louilinn: https://github.com/CirclesUBI/circles-myxogastria/blob/29599c9ae2e28ac7e1c0aaa961f641cc858a769e/src/views/SendConfirm.js#L137

src/utils/findPath.js
louilinn: I will add something like "findTransitiveTransfer returns a maxFlow value when a too large amount is specified as amount. If the amount is in the order of magnitude of actual possible transfers it will return the steps of that transfer instead. We use 10¹⁵ CRC as this unrealistic amount."

src/store/token/actions.js
mikozet: I would log the error. We might think about notification as well so it is a clear point where the actual issue is once it happens..

louilinn:
Try it out in preview :)

src/store/token/actions.js
mikozet: "I will also add some kind of interpretation to the user here"
Is it done somewhere, I do not see?

src/store/token/actions.js
mikozet: I do not see where we catch error from `core.token.updateTransferSteps`

src/store/token/actions.js
mikozet: Is interpretation for the user there?

mikozet (changes requested):
still some comments haven't been addressed

src/store/token/actions.js
louilinn: Yes and it will be improved when working on the task to improve that view

src/store/token/actions.js
```diff
@@ -203,29 +204,105 @@ export function requestUBIPayout(payout) {
...
+      // search complete and no path found
+      (transferError.name === 'TransferError' &&
+        ErrorCodes.TRANSFER_NOT_FOUND) ||
+      // includes errors from attempting transfer with an invalid path
+      transferError.name !== 'TransferError'
+    ) {
+      // update the edges db for trust-adjacent safes
+      try {
+        await core.token.updateTransferSteps(from, to, value, hops);
+      } catch (updateError) {
```
louilinn, line 257: here the update error is caught? Is there something else to be done? @mikozet

src/store/token/actions.js
louilinn: Would it be ok to do this as part of the improve send confirm task?

src/store/token/actions.js
louilinn: I tagged you in a comment with the code

src/store/token/actions.js
mikozet: ok

mikozet (approved):

PR #593, opened 2023-02-06
543 second make notification logic work for news tab

mikozet:


louilinn:
To do: (note to self)
redo 396 branch from main with cherry-picking of relevant commits
force push 396
redo this branch 543 with cherry-picking the last commit 46b649c54b4e7999a9c31f7f707a24d33597845e
force push
PR should be fixed

src/components/NewsFeed.js
```diff
@@ -119,17 +125,18 @@ const NewsFeed = ({ news, isLoading, isMoreAvailable, onLoadMore }) => {
...
 
   return <Box>{newsListElement}</Box>;
 };
 
-const NewsItem = ({ newsItem }) => {
+const NewsItem = ({ newsItem, lastSeenAt }) => {
   const classes = useStyles();
-  const [isExpanded, setIsExpanded] = useState(false);
+
+  const [isExpanded, setIsExpanded] = useState(newsItem.createdAt > lastSeenAt);
```
louilinn, line 139: I think we should probably use date here and have it in a format that can be parsed into display date and be used to make this check. However I will leave this for now and we can figure out the rest after we have the backend

louilinn (approved):
I see some potential clashes with when we add the real backend. I will merge and we can fix that later. Logic works now for the user at least :)

PR #594, opened 2023-02-06
570 trust activity is not opened when entering activity log

### LINKED ISSUE #570: Badge and notification logic + Trust activity is not opened when entering activity log
**Describe the bug**
When a trust event has happened, such as
- trusting or untrusting another account
- being trusted or untrusted by another account
it triggers a badge in the activity log bell.
Upon entering the activity log, the user is referred to the transaction log where nothing has changed. 
The connections log is not opened automatically and shows no badge

**To Reproduce**
1. Enter the activity log first if there are unopened notifications
2. Go to another account and trust or untrust it
3. Go back to the dashboard and wait for the loading wheel to stop spinning on the bell icon
4. See a badge with a 1 (1 new notification)
5. Enter the activity log by clicking the bell
6. Experience the "bug" above
7. Enter the connections tab and see a dot indicating a new notification that wasn't evident before explicitly switching tab.

**Expected behavior**
When entering the activity log I would expect the tab with the highest number to open automatically. If they are even (2 of each for example), start from the left. In the case above this would cause the connections tab to open. If there are notification in multiple tabs I expect the other (non-opened) tab(s) to have badges with numbers of unread notifications. Similar to how badges show in search tabs.

**Device**
All devices.

**Additional context**
Use search view badges as inspiration. Make sure figma design is up to date.

mikozet:
Closes #570

### DIFF
--- a/src/components/ActivityStream.js
+++ b/src/components/ActivityStream.js
@@ -152,8 +152,6 @@ const ActivityStreamList = ({ activities, lastSeenAt, lastUpdatedAt }) => {
     };
   });
 
-  // console.log('ActivityStreamList activities', activities);
-
   if (lastUpdatedAt === 0) {
     return null;
   }
--- a/src/components/ActivityStream.js
+++ b/src/components/ActivityStream.js
@@ -152,6 +152,8 @@ const ActivityStreamList = ({ activities, lastSeenAt, lastUpdatedAt }) => {
     };
   });
 
+  // console.log('ActivityStreamList activities', activities);
+
   if (lastUpdatedAt === 0) {
     return null;
   }
--- a/src/components/ActivityStreamWithTabs.js
+++ b/src/components/ActivityStreamWithTabs.js
@@ -1,6 +1,7 @@
+import { Badge } from '@material-ui/core';
 import PropTypes from 'prop-types';
 import qs from 'qs';
-import React, { Fragment, useEffect, useState } from 'react';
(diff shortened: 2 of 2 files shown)


louilinn (changes requested):
Nice there is improvement :)

However some problems:
Aggregation is broken: 3 transfer notification and 2 trust notifications shows up as 4 in my dashboard bell so some math is wrong.

When both tabs have notifications and you switch between then the badges show up again after you switch back and forth. So there is something wrong with seen logic of tabs. Badge turns of when you switch to the notified tab but reappears when you go back to the other one.
Let's discuss the whole behavior on video sometime?
Maybe it is easier to keep the notifications as unread? but then keep the badge the whole time and don't hide it when opening the tab. I prefer this.

Currently deployed on preview for testing.
Activity log is triggered for sending and receiving circles plus trusting and untrusting plus being trusted and untrusted. Use different combinations of these and then enter the activity log bell icon.

QA:

- [ ] number badge on bell equals total number of new activities
- [ ] number badge on each tab
- [ ] default tab has most notification
- [ ] no notification or same goes to transactions
- [ ] switching between the tabs - verify behavior with Lucia probably

louilinn:
Cherry picked into another PR: https://github.com/CirclesUBI/circles-myxogastria/pull/600/commits/cf2f18767730e5e9651bce42dbc197ea75b85e8f

PR #602, opened 2023-02-10
Make trust button purple for one way trust, regardles of direction

louilinn:
Fixing update of issue #477


louilinn:
I have tested in local that all cases are correct with figma

mikozet (approved):

mikozet:
I noticed a white background showing for less than a second while entering profile view from dashboard. We decided to look on that a bit later

louilinn:
> I noticed a white background showing for less than a second while entering profile view from dashboard. We decided to look on that a bit later

Btw i can't reproduce this in my browser

PR #605, opened 2023-02-11
Update @circles/core version

llunaCreixent:
This core version allows Safes with version `v1.1.1+Circles` to deploy the token contract (`core.token.deploy`).


louilinn (approved):
Thank you!

You have also updated some patch dependencies right?

Issue #531, opened 2022-11-04
Correct Pink Shadow so it does not contain empty space


mikozet:
Subtask of https://github.com/CirclesUBI/circles-myxogastria/issues/396

Hi @ruchajoshi08 
Our current pink shadow is too high and because of that some content might not be clickable.
Could we redesign it in a way that it's height doesn't affect content not being clicked 
Current pink shadow has around 300px height but it could have around 80px or 100px as rest of it looks transparent.
That transparent part causes that we cannot click content below of it..

![Zrzut ekranu 2022-11-4 o 11 25 51](https://user-images.githubusercontent.com/95082155/199952434-597c146a-5577-4d1c-81d4-3dcf9cd6870e.png)
![Zrzut ekranu 2022-11-4 o 11 26 00](https://user-images.githubusercontent.com/95082155/199952439-9309c06a-66c7-4647-89f3-9b8b72170035.png)
![Zrzut ekranu 2022-11-4 o 11 26 06](https://user-images.githubusercontent.com/95082155/199952440-ea60bcce-9aa9-4d0d-bbe2-3232eeda7636.png)

ruchajoshi08:
@mikozet I changed the height to 80px for mobile and 150 for desktop. will it work better now?

mikozet:
> 

@ruchajoshi08 probably yes - could you share a link to design?

ruchajoshi08:
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=4350%3A26250 here is the link

triaslucia:
We still need to fix the shadow. 

-It is too rectangular. We should increase the blurring so that it gets gradually lost, for this it is necessary to increase the height of the surface that covers the shade. 
![Screenshot 2023-02-16 at 09 57 36](https://user-images.githubusercontent.com/38475134/219317153-d300c050-7d9c-464f-a99d-733d7dd88ffb.png)


- There are other issues related to this. So we should treat it a separate thing. 
- It will never be like the design. Should be in HTML, and this makes some changes. 
- We could try to use PNG instead of SVG.
- we are using CSS, but it is not so easy to imitate 100%
- The shadow is now covering other components, and this creates problems This components get inactive as the are under the shadow. 
-

mikozet:
Closing as it is solve here:
https://github.com/CirclesUBI/circles-myxogastria/pull/615

PR #613, opened 2023-02-22
Correct pink shadow so items are clickable

mikozet:
I think it is good for temporary solution as items are clicked but some problems occurs which could be solved by replacing a shadow with version which is less height...
Screenshot from 2023-02-22 10-45-24

### DIFF
--- a/src/components/PinkShadow.js
+++ b/src/components/PinkShadow.js
@@ -24,6 +24,7 @@ const useStyles = makeStyles((theme) => {
       background: theme.custom.gradients.pinkShade,
       opacity: 0,
       transition: 'opacity 0.1s',
+      zIndex: theme.zIndex.layer0,
     },
     isPinkShadow: {
       opacity: '1',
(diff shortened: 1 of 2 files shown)


mikozet:
Also solve this I believe:

https://github.com/CirclesUBI/circles-myxogastria/pull/613

louilinn:
I was able to solve this using pointerEvents set to none:

```
    pinkShadow: {
      width: '100%',
      height: '303px',
      position: 'fixed',
      bottom: '0',
      background: theme.custom.gradients.pinkShade,
      opacity: 0,
      transition: 'opacity 0.1s',
      pointerEvents: 'none',
    },
```

https://stackoverflow.com/questions/65476127/the-parent-box-shadow-above-the-clickable-button

So what I think we should do now is to apply that styling in PinkShadow.js component and make the avatars etc have normal z-level so that they are also behind the shadow visually

louilinn:
I'll take it over since we should branch this off main instead

louilinn:
Closed in favor of: https://github.com/CirclesUBI/circles-myxogastria/pull/615

louilinn:
> Also solve this I believe:
> 
> #613

I think you pasted the wrong link? You are referencing this same PR?

mikozet:
> I think you pasted the wrong link? You are referencing this same PR?

Yes I was talking about this one 
https://github.com/CirclesUBI/circles-myxogastria/issues/531

PR #615, opened 2023-02-26
Let pink shadow allow for clicking on components behind it

louilinn:
I was able to solve this using pointerEvents set to none:

```
    pinkShadow: {
      width: '100%',
      height: '303px',
      position: 'fixed',
      bottom: '0',
      background: theme.custom.gradients.pinkShade,
      opacity: 0,
      transition: 'opacity 0.1s',
      pointerEvents: 'none',
    },
```

https://stackoverflow.com/questions/65476127/the-parent-box-shadow-above-the-clickable-button

So what I think we should do now is to apply that styling in PinkShadow.js component and make the avatars etc have normal z-level so that they are also behind the shadow visually

- [x] remove pointer events in pink shadow
- [x] redo z-level  for avatars


mikozet (approved):

mikozet:
well done! Nice solution!

Issue #611, opened 2023-02-15
Loading screen forever after updateTransferSteps


JacqueGM:
**Describe the bug**
After the updateTransferSteps - the transactions are not being retried therefore the screen remains in a loading page
**To Reproduce**
1. Go to circles.garde
2. Click on send and open network tab
3. Wait until it finished the whole sending process.
4. Observe that after the update steps the transaction are not retried ending up in a forever loading screen

**Expected behavior**
After update steps the transaction should be retried and a message should be shown on what is happening

**Screenshots**
<img width="1414" alt="Screenshot 2023-02-15 at 16 12 56" src="https://user-images.githubusercontent.com/55389166/219068069-29e784ef-e215-4b91-91c5-c0dae69810ed.png">

**Desktop-Computer (please complete the following information):**
 - OS: MacOS
 - Browser Chrome

louilinn:
waiting for replacement of pathfinder service first

mikozet:
This PR might be helpfull for this task: https://github.com/CirclesUBI/circles-myxogastria/pull/627
As it solves issue in coment described here:
https://github.com/CirclesUBI/circles-myxogastria/issues/589#issuecomment-1462356650

juanenrisley:
As seen with @JacqueGM this issue is caused by #644. We tested in https://preview-dev.circles.garden/ and does not occur anymore.

PR #627, opened 2023-03-14
589 transaction of high amount are being reverted

### LINKED ISSUE #589: Transaction of high amount are being reverted
**Describe the bug**
When sending high amounts (>8k) the transactions are reverted but the message says transaction done.

**To Reproduce**
1. Go to transfer CRC
2. Click on  send
3. Wait until transaction takes place

**Expected behavior**
Transactions should be done and if they are not done a proper message should be returned, but not a success message
**Screenshots**
[](url)

**Desktop-Computer (please complete the following information):**
 - Chorme

[
Screenshot 2023-01-30 at 14 15 47

Screenshot 2023-01-30 at 14 57 52
](url)

mikozet:


louilinn:
@mikozet will file a new issue that better describes the bug in resloved in this PR
In this PR he will add a description explaining why this is a good solution, how behaviour changes and why

src/views/SendConfirm.js
```diff
@@ -117,8 +118,27 @@ const SendConfirm = () => {
...
     try {
-      await dispatch(transfer(address, amount, paymentNote));
+      await dispatch(
+        transfer(
+          address,
+          amount,
+          paymentNote,
+          PATHFINDER_HOPS_DEFAULT,
+          PATHFINDER_HOPS_DEFAULT + 1,
+          additionalAction,
```
mikozet, line 139: the idea behind this is it gives us flexibility what to do in case of for example endless spinner. We define that additional action where transfer function gets dispatched. Our loading spinners have state in various components so thus that should be closed from there

louilinn:
What was perhaps fixed by this PR (issue #611) is resolved by https://github.com/CirclesUBI/circles-myxogastria/pull/648

PR #642, opened 2023-03-30
Don't call getTransactionReceipt if txHash is not defined

### LINKED ISSUE #641: Invalid params - error in console
Worked on in bug session 28/3-23

Sentry ids:
MYXOGASTRIA-PRODUCTION-EBM
MYXOGASTRIA-PRODUCTION-FAC
Top cause of error in Sentry.

Source: https://github.com/CirclesUBI/circles-myxogastria/blob/551cddf24ab687469c2ac625f2fd221a47beef00/src/store/activity/actions.js#L70
Does not directly effect users. The activity status could not be verified. It retries again later anyway when state changes (method runs in hook). But it spams console and Sentry.

hypothesis: 
1. web3 method is responding with an error that is not handled
2. the web socket has disconnected for some reason

Running a scripted with invalid transaction hashes to proved theory 1. We need to validate txHash is defined (unlikely to be defined but incorrect).

Documentation claims `getTransactionReceipt` returns null when txHash is incorrect but sometime / often it triggers errors. Running the script it does so for example.

The error is hard to reproduce in production. We don't really know why this is. 
We try validating and see if errors go down in Sentry as step 1.

louilinn:
Closes: https://github.com/CirclesUBI/circles-myxogastria/issues/641

As a first step verify txHash is defined.
If this does not help we can catch the error or validate further.

### DIFF
--- a/src/store/activity/actions.js
+++ b/src/store/activity/actions.js
@@ -68,7 +68,9 @@ export function checkPendingActivities() {
 
     for await (const category of CATEGORIES) {
       activity.categories[category].activities.forEach(async (activity) => {
-        if (!activity.isPending) {
+        // We only need check pending activities and activities without
+        // txHash are not ready to be checked.
+        if (!activity.isPending || !activity.txHash) {
           return;
         }
 


mikozet (approved):

juanenrisley (approved):
👍🏿

PR #649, opened 2023-04-03
633 export balance UI for first stage visual part and popup functionality with date selection

### LINKED ISSUE #633: Export Balance UI for first stage - visual part and popup functionality with date selection
Design material:
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=7523-11295&t=VtAkrRww9liYOQpk-0#394798365 - FIRST STAGE

Create the graphical interface and allow for date selection.
The download behavior is set to the export button in design.

This task does not involve manipulating downloadCsvStatement or file statement. It is only the UX parts.
(see task issue #623 for next functionality task)

louilinn:
Closes #633


mikozet:
I didn't notice that DateRangePicker requires to use commercial license of MUI :( more info in link
https://github.com/mui/mui-x
"Pro Plan
The Pro version of MUI X expands on the features of the free version with more advanced capabilities such as multi-filtering, multi-sorting, column resizing and column pinning for the data grid; as well as the date range picker component.

The Pro version is available under a commercial license—visit the Pricing page for details."

Because of that my proposition is to use two DatePickers MUI components.
How it will work:
We set date in 'From' input and in 'To' input it automatically adjust minDate so we cannot choose date before 'From' - I already implemented that logic..
Pros:
We still have MUI design, it gives good mobile experience as well, we do not to install additional libraries

I tried also to find some alternatives:
I checked two libraries:
1. https://reactdatepicker.com/#example-date-range (I didn't like it)
2. https://react-dates.github.io/react-dates/?path=/story/daterangepicker-drp--default (this one is/was used by Airnb)

I implemented second solution but there were some drawbacks:
We need to install additional libraries. Styling of inputs is not that easy and I am not sure if we would be able to replicate MUI looking input which we currently have (I doubt it as html structure is different and text is presented as input attributes not additional html elements)
Another thing which I discovered is that on mobile calendar was cut and there was no scroll (in our case, probably due to other styles in our app) so that would also require some adjustments.   

To some up I decided two go with two DatePickers (currently inputs still needs styling)

louilinn:
We have merge to rebase the export-balances branch to get the MUI updates

PR #656, opened 2023-04-12
Update actions versions and use setup-node for caching

llunaCreixent:
Using the `.nvm` file works. However, caching the dependencies is not working. I suggest keep using the `actions/cache` action.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,26 +8,14 @@ jobs:
 
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v2
-
-      - name: Read node version from .nvmrc
-        id: nvmrc
-        run: echo ::set-output name=NODE_VERSION::$(cat .nvmrc)
+        uses: actions/checkout@v3
 
       - name: Setup node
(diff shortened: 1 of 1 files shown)


llunaCreixent:
Check in the [Action](https://github.com/CirclesUBI/circles-myxogastria/actions/runs/4675900375/jobs/8281544972) the following log: `npm cache is not found`

llunaCreixent:
Do you know how to cache the dependencies without using the `actions/cache` action? [Here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) it says it's not necessary since you can configure the cache with `setup-node` action. We have 2 options:
1. Keep using `actions/cache`
2. Learns how to configure the cache through `setup-node` action

llunaCreixent:
Perfect!
We accomplished option 2 (configure the cache through `setup-node` action).
You can check how the cache is found here: https://github.com/CirclesUBI/circles-myxogastria/actions/runs/4678146854/jobs/8286524579

All looks good can be reused for other repos

juanenrisley (approved):
👍🏿

louilinn (approved):

Issue #623, opened 2023-03-07
Export Balance for a customized date interval


louilinn:
Design material:
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=7523-11295&t=VtAkrRww9liYOQpk-0#394798365 - FIRST STAGE

See last image/frame on the right as a reference to this stage of developing the feature

This task entails:
- adjusting the the export logic to allow for timespans

louilinn:
@triaslucia  lets discuss dates here

## Regarding time intervals and reasons for them:
### One year limit:
Reason 1:
To do the export we need to fetch all transactions to the starting date and calculate old balances based on all transactions after. The longer back in time we go the more calculations we need to make and the more data we need to fetch. It is the starting date not the time interval length that determines the complexity of calculating end and start balance.

Reason 2:
When asking for the transactions from the Graph service we have to provide a maximum number of transactions. This is currently set as 1000. I don't know how much higher we can set it. I could experiment but I started out here.

It means also that right now **exports will not be complete more than 1000 transactions back in time**. This is the only **know flaw** atm of the export. Most users don't have more than 1000 in a year so I set the max start date one year to limit the expectation on the export.  There isn't away of knowing the earliest possible start date before exporting so I set a generic limit, to minimize this confusion for most users. 

### Max interval 3 months:
At the moment exporting is a bit heavy because the backend is not optimized for getting payment notes in bulk (many at the same time).
To avoid the user experiencing the export time as too long I suggest we limit the time interval to 3 months. Again time/complexity is based on the number of transactions, not time, but they correlate.

It is also a way to prevent users from exporting unnecessary big amounts of data for only testing purposes, rather then actually needing it. If the usecase to export more than a quarter at a time is common we can discuss to change it.

### General
What I try to limit is the number of transactions to export and fetch, I do this by restricting dates, but transaction intensity is different between users. So we have to explore this more, how to deal with limits in fetching and so on.

triaslucia:
Thanks! 
All sound good to me. WE can go this way, and in a future revise, if needed!

PR #661, opened 2023-04-19
623 Use date interval provided for export wallet statement

louilinn:
Closes: #623 

- [x] filter fetched transactions based on provided dates
- [x] fetch more transactions and based on date
- [x] refine date selection limitations (max 3 months in duration, max 1 year back)
- [x] default dates set to last month

Remaining that we can finish in another task/PR
- [ ] unit test
- [ ] error handling
- [ ] solve problem when fetching Payment notes for a large number of transactions

### DIFF
--- a/src/utils/fileExports.js
+++ b/src/utils/fileExports.js
@@ -1,5 +1,6 @@
 import { crcToTc } from '@circles/timecircles';
 import fileDownload from 'js-file-download';
+import _ from 'lodash';
 import { DateTime } from 'luxon';
 
 import core from '~/services/core';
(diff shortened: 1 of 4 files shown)


louilinn:
@mikozet please review before you work on export balances
and preferably we merge this before you work more on the UX parts

src/utils/fileExports.js
```diff
@@ -1,5 +1,6 @@
 import { crcToTc } from '@circles/timecircles';
 import fileDownload from 'js-file-download';
+import _ from 'lodash';
```
mikozet: can we import just partition function instead of whole library?

mikozet (approved):

PR #663, opened 2023-04-20
Optimizing api calls for payment notes in wallet statement export

louilinn:
- [x] call getPaymentNote / resolveTxHash in sequence using observables, so that the api does not crash when making many requests
- [x] use object for transaction instead of array

coded with @juanenrisley :star2: thanks for the help!

### DIFF
--- a/src/utils/fileExports.js
+++ b/src/utils/fileExports.js
@@ -96,15 +96,23 @@ const formatTransactions = async (transactions, safeAddress) => {
       false, // important to round normal, not down to get even numbers
     );
 
-    return [
+    return {
       // TODO object
-      formatDate(date),
-      'placeholder name',
-      otherSafeAddress,
-      '-',
-      valueSign.concat(valueInTimeCircles),
-      '',
-    ];
+      date: formatDate(date),
+      //name: 'placeholder name',
+      otherSafe: otherSafeAddress,
+      //paymentNote: '-',
+      amount: valueSign.concat(valueInTimeCircles),
+    };
+    // [
+    //   // TODO object
+    //   formatDate(date),
+    //   'placeholder name',
+    //   otherSafeAddress,
+    //   '-',
+    //   valueSign.concat(valueInTimeCircles),
+    //   '',
+    // ];
   });
 
   const notes = await getPaymentNotes(transactions);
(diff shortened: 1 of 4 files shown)


src/utils/fileExports.js
```diff
@@ -16,6 +18,18 @@ const { ActivityFilterTypes } = core.activity;
 
 const formatDate = (dateTime) => dateTime.toFormat(`dd.LL.yyyy`).toString();
 
+const loadPaymentNote = (txHash) => {
+  return from(resolveTxHash(txHash).then((note) => note || '-'));
+};
+
+const getPaymentNotes = (transactions) => {
+  const paymentNotes$ = from(transactions).pipe(
+    mergeMap((tx) => loadPaymentNote(tx.txHash), 1),
```
mikozet, line 27: it would be worth to try with different values like 2,5 or 10
`mergeMap((tx) => loadPaymentNote(tx.txHash), 2)` as an example, it should increase perforamance and potentionally still not crash

mikozet:
It behaves a bit strange for me at the beginning. 
First export is always longer and includes some errors...
It doesn't happend for the second time...
Screenshot from 2023-04-26 09-12-46

src/utils/fileExports.js
juanenrisley: That's a nice idea.

juanenrisley (approved):
👍🏿

louilinn:
> It behaves a bit strange for me at the beginning. 
> 
> First export is always longer and includes some errors...
> 
> It doesn't happend for the second time...
>
> 

It takes longer because each payment note is requested in sequence rather than all at once. This is to prevent overloading the api. 

The number of errors are identical to before (for same wallet and same period). The api responds with 404 when the tx hash doesnt exist in the db, which it doesnt when payment note is enpty / non existent. I will file an issue in api to change this behaviour and also to request many payment notes at once. 

You can see the same 404 error if you go to avtivity log and open a transfer that doesn't have a payment note.

src/utils/fileExports.js
louilinn: I will try this :)

mikozet:
> It takes longer because each payment note is requested in sequence rather than all at once. This is to prevent overloading the api.
> 
> The number of errors are identical to before (for same wallet and same period). The api responds with 404 when the tx hash doesnt exist in the db, which it doesnt when payment note is enpty / non existent. I will file an issue in api to change this behaviour and also to request many payment notes at once.
> 
> You can see the same 404 error if you go to avtivity log and open a transfer that doesn't have a payment note.

but did you notice that at first time of export there are errors and it takes longer to load and when you do export second time 
 there are no errors and it is faster (without changing date range)

mikozet:
note from meeting: it's probably because it is cached

mikozet (approved):

src/utils/fileExports.js
```diff
@@ -190,8 +205,7 @@ export async function downloadCsvStatement(
     transactions,
     (tx) => tx.date < endDate,
   );
-  // TODO change to filter for txsInPeriod
-  const [txsInPeriod] = partition(txsBeforeEnd, (tx) => tx.date < startDate);
```
louilinn, line 194: this broke the whole fuctionality @mikozet but it is fixed in this PR again when merging.
(you basically removed all transactions, so none were included in #668)

PR #668, opened 2023-04-26
664 export balances notifications and loading

mikozet:
- [x] Add errors notification
- [x] Add confirmation notification


louilinn (approved):
I think we can improve error handling a bit but this is good enough for now

PR #669, opened 2023-04-26
Add error state for DateInput

mikozet:
- [x] Add error state for date input
- [x] Add text info error for date input
- [x] Disable 'Export Statement' button when there is an error in input

### DIFF
--- a/src/components/DialogExportStatement.js
+++ b/src/components/DialogExportStatement.js
@@ -148,10 +148,10 @@ const DialogExportStatement = ({ dialogOpen, onCloseHandler }) => {
     switch (errorToInput) {
       case 'maxDate':
       case 'minDate': {
-        return 'Select a correct date range';
+        return translate('ExportStatement.exportHintRange');
       }
       case 'invalidDate': {
-        return 'Your date is not valid';
+        return translate('ExportStatement.exportHintDate');
       }
       default: {
         return '';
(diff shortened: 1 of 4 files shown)


src/components/DialogExportStatement.js
```diff
@@ -132,13 +134,45 @@ const DialogExportStatement = ({ dialogOpen, onCloseHandler }) => {
...
+        return '';
+      }
+    }
+  }, [errorToInput]);
+
+  const errorMessageFrom = useMemo(() => {
+    switch (errorFromInput) {
+      case 'maxDate':
+      case 'minDate': {
+        return 'Select a correct date range';
```
louilinn: Replace with "Invalid range"

src/components/DialogExportStatement.js
```diff
@@ -132,13 +134,45 @@ const DialogExportStatement = ({ dialogOpen, onCloseHandler }) => {
...
+      }
+      case 'invalidDate': {
+        return 'Your date is not valid';
```
louilinn: Replace with 'Invalid date'

src/components/DialogExportStatement.js
```diff
@@ -132,13 +134,45 @@ const DialogExportStatement = ({ dialogOpen, onCloseHandler }) => {
...
+  }, [startDate, endDate, errorFromInput, errorToInput]);
+
+  const errorMessageTo = useMemo(() => {
+    switch (errorToInput) {
+      case 'maxDate':
+      case 'minDate': {
+        return 'Select a correct date range';
+      }
+      case 'invalidDate': {
+        return 'Your date is not valid';
```
louilinn: See comments of new messages. We can keep the logic for now. But the two messages needs to be in en.json, not defined here!

louilinn (changes requested):

louilinn (approved):

PR #675, opened 2023-05-02
188 export balances error state

mikozet:


louilinn (changes requested):
Something with the dates is broken

src/components/DialogExportStatement.js
```diff
@@ -201,6 +204,7 @@ const DialogExportStatement = ({ dialogOpen, onCloseHandler }) => {
           </Typography>
           <Box className={classes.dateInputsContainer}>
             <DatePicker
+              format="dd/MM/yyyy"
```
louilinn, line 214: maybe it needs to be DD/MM/YYYY ? possibly. I havent checked docs yet

mikozet: Hey I checked again and I think it was a bug in preview mode in my libre office. Once I open file it actually shows all records
preview mode:
Screenshot from 2023-05-02 16-32-17

and here is actual file:
Screenshot from 2023-05-02 16-32-31

juanenrisley (approved):
👍🏿 Everything is allright

Issue #678, opened 2023-05-04
Commas in payment notes break 'Amount_in_circles' column in exported CSV


juanenrisley:
Having commas in a payment note text adds more columns to the CSV as the syntax is comma separated for column division. This breaks the _Amount_in_circles_ column as it holds text values and not the amount, that is left behind in a new column.

For example, this payment note was: `donation, don't send back please`

![Screenshot from 2023-05-04 08-50-08.png](https://images.zenhubusercontent.com/63e50e3f472d0b732cddc904/0660c270-7305-445b-a6d6-560eb19dcd45)

llunaCreixent:
@mikozet @louilinn do you wanna close this issue?

mikozet:
yes I believe it is done, I am closing

louilinn:
Resolved by https://github.com/CirclesUBI/circles-myxogastria/pull/681

PR #681, opened 2023-05-08
Fix commas in csv with double quotation marks around payment note

louilinn:
Solves #678


mikozet (approved):

louilinn:
for reference:
Screenshot from 2023-05-09 10-26-28
these weird messages are handled correctly

PR #684, opened 2023-05-10
Fix state to distinguish invalid and downloading

louilinn:
Before, downloading / loading text was showing when date was invalid. There are two cases for disabled button. These are now separated.


juanenrisley (approved):
👍🏿

PR #687, opened 2023-05-25
Update Typography

### LINKED ISSUE #482: Refactor text styles
We should refactor all text styles on our wallet. 

**Design Material** https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?node-id=4673%3A9745

mikozet:


mikozet:
- [x] Create font styles in themes / text variants
- [x] Apply variants to all places we have text
- [x] Delete old and redundant css for text
- [x] Update Heading 4
- [x] Update Heading 5
- [x] Update Input Transfer Circle font size 
- [x] Update title for steps in stepper counter   
- [x] We temporary disable language switcher since languages are not there but should come soon

src/views/Error.js
```diff
@@ -36,26 +36,44 @@ const CriticalError = () => {
             : translate('CriticalError.bodyCriticalErrorTryAgain')}
         </Typography>
         <AppNote messageVersion="error" />
-        {app.errorMessage && (
+        {!app.errorMessage && (
```
juanenrisley: Is this a mistake? Does make sense here to check the condition `!app.errorMessage`?

src/styles/fonts.js
```diff
@@ -16,8 +16,17 @@ export const fontFamily = 'Noto Sans';
 export const fontWeightLight = 300;
 export const fontWeightRegular = 400;
 export const fontWeightMedium = 500;
+export const fontWeightMediumBold = 600;
 export const fontWeightBold = 700;
 
+export const fontSizeSmallest = '10px';
```
juanenrisley: Maybe here we could define all font sizes as just numbers not as strings with 'px' so they would be used in other cases apart from font sizes like in CircularProgress's size or CirclesLogoSVG's width e.g.: https://github.com/CirclesUBI/circles-myxogastria/blob/3a40d077ce200fdeecd65064141a2a36eda747b9/src/components/TransferInfoCard.js#LL64C18-L64C32

src/views/Error.js
mikozet: thx for noticing it

src/styles/fonts.js
mikozet: done

juanenrisley (dismissed):
👍🏿 Well done!

louilinn (approved):

PR #689, opened 2023-06-08
Update header in login view once scrolled

### LINKED ISSUE #635: Header in login view is white when scrolled instead of turquoise
Header is the wrong color and header icons have wrong color.
Should be the same as personal dashboard.

Screenshot of incorrect design
Screenshot from 2023-03-30 16-34-14
Screenshot from 2023-03-30 16-34-01

mikozet:


src/views/Login.js
```diff
@@ -101,7 +101,11 @@ const Login = () => {
 
   return (
     <BackgroundCurved gradient="turquoise">
-      <Header>
+      <Header
+        hasWhiteIcons
+        isOrganization={false}
+        useSpecialWithColorOnScroll={true}
```
louilinn, line 107: technically you can skip the true part but ok :)

louilinn (approved):

PR #690, opened 2023-06-09
Disable notification spam during not allowing camera access in browser

### LINKED ISSUE #639: Not allowing camera access in browser for editing avatar triggers notification spam
**Describe the bug**
Not allowing camera access in browser for editing avatar triggers notification spam

**To Reproduce**
Steps to reproduce the behavior: [in DESKTOP BROWSER]
1. Go to edit profile through floating menu
2. Click on the avatar to update it
3. Choose camera
4. When browser asks if circles.garden can access camera, chose deny
5. See notifiation spam, which does not stop even if you change view unless you reload the page.

**Expected behavior**
One of these notifications persisting until closed and then nothing else.

**Screenshots**
Screenshot from 2023-03-28 14-16-16

**Desktop-Computer (please complete the following information):**
 - OS: Debian 11
 - Browser: Chromium

mikozet:


louilinn:
Hi @mikozet,
Can you describe what was the problem and what is the new behaviour? Does it still through the error but not in a spammy way?

louilinn:
Hi @mikozet,
Can you describe what was the problem and what is the new behaviour? Does it still through the error but not in a spammy way?

mikozet:
> Hi @mikozet, Can you describe what was the problem and what is the new behaviour? Does it still through the error but not in a spammy way?

Hey, sure.
Problem was that when there was an error onError function got triggered which displayed notification. But that onError function also caused hook to rerun and throwing an onError again - creating an endless loop

mikozet:
Yes, the error is now trigerred but not in a spammy way

louilinn (approved):
Thanks Miko! :) :star:

PR #692, opened 2023-06-16
Add error notification if user token is not deployed or failed to deploy

### LINKED ISSUE #645: Undeployed account gets zero address as token address and the error should be treated differently
[MYXOGASTRIA-PRODUCTION-4H8](https://circles-ubi.sentry.io/issues/2253587336/?project=5446133&query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=30d&stream_index=4)

When an account is not yet verified (deployed) token returned is the zero address

This should be reflected in the [core method documentation](https://github.com/CirclesUBI/circles-core/blob/0c8479cea08029f5540ba1c3aa42adbb214f0291/src/token.js#L342)

[Myxogastria code](https://github.com/CirclesUBI/circles-myxogastria/blob/551cddf24ab687469c2ac625f2fd221a47beef00/src/store/token/actions.js#L107)

We should catch the error thrown for Zero address and display an error/warning notification that this user's token has not yet been deployed or failed to deploy. - Ask product for message and tell them about the error and when it occurs.

Catch in [actions here](https://github.com/CirclesUBI/circles-myxogastria/blob/551cddf24ab687469c2ac625f2fd221a47beef00/src/store/app/actions.js#L90) and dispatch notification.

mikozet:


src/store/token/actions.js
```diff
@@ -107,6 +109,12 @@ export function checkTokenState() {
       const address = await core.token.getAddress(safe.currentAccount);
 
       if (address === ZERO_ADDRESS) {
+        dispatch(
```
louilinn: What about the dispatch on line 129? I think that is the better place to dispatch the error message. We should rather review that current on?
@juanenrisley what do you think?

juanenrisley: Yeah i agree. Should be better to reuse the dispatch function and just parametrize the data we are going to dispatch rather than calling dispatch multiple times everywhere.

mikozet: Dispatch in line 129 is called once there is an error from catch block. We do not receive an error from core when `address === ZERO_ADDRESS`. So dispatch placed over there will not get triggered.

juanenrisley: An error is being throwed in line 118 so the catch will be triggered with it anyway: https://github.com/CirclesUBI/circles-myxogastria/blob/c708df4f9e394d34a343e81ac8a95f8b79e60e5f/src/store/token/actions.js#L118

mikozet: true, corrected

juanenrisley (dismissed):
👍🏿 Alrighty!!!

juanenrisley (dismissed):
👍🏿

louilinn (approved):

PR #696, opened 2023-06-20
Upgrade node to version 16

louilinn:


package.json
juanenrisley, line 1: For me, the only thing is missing is setting the "engines" property in the package.json like here: https://github.com/CirclesUBI/circles-core/blob/main/package.json#L19

louilinn: Thanks! :) :+1:

juanenrisley (approved):
👍🏿 Alrighty!

louilinn:
Pending for next release. To be squashed and merged.

.nvmrc
```diff
@@ -1 +1,2 @@
-14
+16
+engine-strict=true
```
mikozet: 1. Shouldn't it be in .npmrc file ?
2. I'm not sure if I like it I have some doubts - I think it is not necessary to have it and 'engines' in package.json is potentionally enough. It might bring more troubles than solutions and maybe requires more tests? For example did you try some tests how it works if some 3rd party package is not compatible with node 16 but higher one or lower one? Even if it is true everything might still work...  It will probably give us an Error instead of Warning once some even 3rd party package is not compatible with node 16 and we do not have control over it...
referring to the docs:
https://docs.npmjs.com/cli/v9/using-npm/config#engine-strict

package.json
```diff
@@ -15,6 +15,9 @@
     "type": "git",
     "url": "git://github.com/CirclesUBI/circles-myxogastria.git"
   },
+  "engines": {
```
mikozet, line 18: this is nice🌻

.nvmrc
juanenrisley: As the docs says "This can be overridden by setting the --force flag.". To be honest, If there is a safe hatch i dont know if it's a real problem.

PR #701, opened 2023-06-26
Add UI for transaction filters

### LINKED ISSUE #558: Transactions spliting Received-sent
**People should be able to see a split view of their transactions.**
while standing on the Activity Log screen on the "transactions" tab, people should be able to choose to display only the received or only the sent transaction - apart from the already existing overall view order by date. This is helpful for business to be able to track better the payments the receive. 

**Design Material**
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Circles-myxogastria?type=design&node-id=9427-16806

mikozet:


src/components/ActivityStream.js
```diff
@@ -167,6 +180,22 @@ const ActivityStreamList = ({ activities, lastSeenAt, lastUpdatedAt }) => {
             return acc;
           }
 
+          const info =
+            data.from === safeAddress
+              ? { prefix: '-', type: 'SENT' }
+              : { prefix: '+', type: 'RECEIVED' };
+
```
juanenrisley, line 187: Both `if` statements could be in the same one as both are used for returning the same value

louilinn: With an OR || between statements on line 189-190 and line 195, improve readability :)

louilinn (changes requested):
Small code comment, otherwise great. 

One consideration is that when filtering and there is only one or maybe no received transaction the log will have 1 or no items and load more, rather than 10. We could trigger another call but I think we should only act on this behaviour is product brings it up. I see this as good enough :)

louilinn:
Another thing we noticed: When switching to Received or Sent the title of the drop down should change accordingly

mikozet:
> Another thing we noticed: When switching to Received or Sent the title of the drop down should change accordingly

it is corrected now

juanenrisley:
@mikozet i see that the tests werent successful, maybe is something temporary. Could they be relaunched manually?

juanenrisley (approved):
👍🏿

louilinn (approved):

PR #704, opened 2023-06-29
Update logo on Welcome and Login page

### LINKED ISSUE #703: Change illustration on Welcome and log in page
**Update the  default profile picture on log in page**
 
_Log In page_ 

remove the old Circles illustration, with people on top. And change for the Circles Coop one, with Yellow background. 

https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Product-and-UX-design-circles.garden-(Circles-myxogastria)?type=design&node-id=9774-13062&mode=design&t=pwLWMxbM5bKJe7Dd-0

mikozet:


louilinn (commented):
I think now we are creating legacy code. Why keep the old logos? 
Doesn't it make more sense to replace the old svg files of logo with new logo?

My suggestion: 
If we are changing the log we should replace the circlesLogoSVG file
We should keep the Logo component.
We should not add the logo to icons file as it is not an icon, its the logo

louilinn:
Let's put this on hold until we have confirmed if logo is changing fully or not. I've asked product in the issue

triaslucia:
Things are not so organized as far as IO can see. There are currently 2 images existing. How this is named on the code, no idea. But for the outside we have two different “logo” images in the App.

_Colourful C, turned down with "gang" on top_
-Login
-Welcome 

_Circles logo (colorful C- in correct position)_ 
-Favicon / tab icon
-Max transfer amount icon
-CirclesUBI payout "avatar" in activity log
-Saving the webapp to home screen on phone creates an icon image
-Validation loading screen
-Preview image in links

**Necessity:** we need to get rid of the Colorful C, turned down with "gang" on top. ASAP. 
**Proposal:** clear up code/Assets, eliminate the Colorful C, turned down with "gang" on top, and use for now the . Circles logo (colorful C- in correct position) for all situations as 
**this is priority 1, in order to move forward with the guidelines**

src/components/Logo.js
```diff
@@ -48,8 +24,8 @@ const Logo = ({ isWithGang, size = 'default' }) => {
 };
 
 Logo.propTypes = {
-  isWithGang: PropTypes.bool,
   size: PropTypes.string,
+  type: PropTypes.string,
```
juanenrisley: I guess this property does not make sense anymore as it is not used within the component.

src/components/Logo.js
```diff
@@ -2,41 +2,17 @@ import { Box } from '@mui/material';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import CirclesLogoGroupSVG from '%/images/logo-group.svg';
 import CirclesLogoSVG from '%/images/logo.svg';
 
-const Logo = ({ isWithGang, size = 'default' }) => {
+const Logo = ({ size = 'default' }) => {
   const sizes = {
```
juanenrisley: This constant could be extracted and declared outside the functional component. There is not need to let it within the function to be redeclared everytime the component is instantiated since it does not depend from component props and is just sizes definition.

src/components/Logo.js
mikozet: corrected

src/components/Logo.js
mikozet: it actually depends on line 18 and 20:
`height={sizes[size]}`
`width={sizes[size]}`

juanenrisley: I meant that the variable's value does not depend on anything. The `sizes` object is a constant always with the same value so can be declared outside the `Logo` component.

mikozet: ok, I got it now :). Thx I corrected it :)

juanenrisley (approved):
👍🏿

louilinn (approved):
:star2:

Issue #272, opened 2021-12-16
Load more button on Activity tab doesn't show any new items after clicking it


mikozet:
New users have very little activities. On activity tab there is `load more` button which after clicking doesn't load any new items. What is expected is that if there are no new items to show to user we shouldn't display `load more` button

louilinn:
Lucia agrees to remove load more when not more is available

mikozet:
In current behaviour Load more button will be showed incorrectly only when remeaning items of user is equal to page size. So for example if we display 10 items per page and user have 10 total items Load more button will be showed incorectly (as there is no more items to show). If total items are 20 at the second page Load more button will be showed incorrectly. In other cases it disappears correctly.

Problems occurs in Activity and Mutual activities
for example:
/profile/0x559379752eE03b8C4644B2B353fecdd836e41A97
/activities?category=transfers

louilinn:
Resolved :)

PR #713, opened 2023-07-20
102 show user related activities in profile

### LINKED ISSUE #102: Show user related activities in Profile
https://www.figma.com/file/ZTqF54s6CSHTwAxp0U0JjQ/Product-and-UX-design-circles.garden-(Circles-myxogastria)?type=design&node-id=7523-11365&mode=design&t=yPVTJsS38AsNIaUw-0 from Lucia

mikozet:
- [x] refresh mutual activities when new activity just happened
- [x] keep mutual activities when changing tabs inside of the profile or when going out but coming back to the same user
- [x] fix loading indicator when we have loading state of a query - display loader
- [x] corrected behaviour - start on activities with loading state instead of empty item state as we do not know if there are items to load or not
- [x] Fix loading more button behaviour on activities and mutual activities. Add comment to task https://github.com/CirclesUBI/circles-myxogastria/issues/272 as it's scope got changed


src/components/ProfileContentActivity.js
```diff
@@ -0,0 +1,79 @@
...
+        dispatch({
+          type: ActionTypes.ACTIVITIES_MUTUAL_ADDRESS_UPDATE,
+          meta: {
+            mutualAddress: address,
+          },
+        }),
+        dispatch(loadMoreActivitiesMutual(address, { fromOffsetZero: true })),
+      ]);
+    }
+  }, [address, mutualAddress, dispatch]);
```
juanenrisley: I think is best to not declare non changing functions as `dispatch` as dependencies in the `useEffect` hook

src/components/ActivityStream.js
```diff
@@ -194,18 +208,28 @@ const ActivityStreamList = ({
           }
 
           const isSeen =
-            DateTime.fromISO(lastSeenAt) > DateTime.fromISO(createdAt);
+            lastSeenAt && createdAt
+              ? DateTime.fromISO(lastSeenAt) > DateTime.fromISO(createdAt)
+              : true;
+
+          const key = hash ? hash : transactionHash;
```
juanenrisley: This kind of declaration is the same as `const key = hash || transactionHash;`. Here are 3 kind of these definitions

src/components/ProfileContentActivity.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const currentTime = DateTime.now().toISO();
+
+  const handleLoadMore = () => {
+    dispatch(loadMoreActivitiesMutual(address));
+  };
+
+  return (
+    <>
```
juanenrisley: This empty tag wrapper (`<></>`) could be removed.

juanenrisley (dismissed):
Everything is fine. All my comments are kinda optional, could be omitted if you do not see them as worth changing.

src/components/ProfileContentActivity.js
mikozet: corrected

src/components/ProfileContentActivity.js
mikozet: corrected

src/components/ActivityStream.js
mikozet: corrected

juanenrisley (dismissed):
👍🏿

src/components/ActivityStream.js
```diff
@@ -194,18 +208,27 @@ const ActivityStreamList = ({
...
               <ActivityStreamItem
-                createdAt={createdAt}
+                createdAt={createdAtDate}
                 data={data}
                 isPending={isPending}
                 isSeen={isSeen}
                 prefix={info.prefix}
                 safeAddress={safeAddress}
-                txHash={txHash}
+                txHash={txHashUpdtated}
```
louilinn: typo

src/components/ActivityStream.js
```diff
@@ -438,11 +461,11 @@ ActivityStreamList.propTypes = {
 };
 
 ActivityStreamItem.propTypes = {
-  createdAt: PropTypes.string.isRequired,
+  createdAt: PropTypes.string,
   data: PropTypes.object.isRequired,
-  isPending: PropTypes.bool.isRequired,
+  isPending: PropTypes.bool,
```
louilinn, line 466: could you just breifly explain why createdAt and isPending should not be required while prefix should?

louilinn: All stream items dont have prefix or?

src/components/ButtonAction.js
```diff
@@ -14,6 +14,7 @@ const useStyles = makeStyles((theme) => ({
     bottom: theme.spacing(2.25),
     right: theme.spacing(2.25),
     background: 'transparent',
+    zIndex: theme.zIndex.layer2,
```
louilinn, line 17: have you checked through the app that this layer change doesn't mess up any other view since it is a commonly used component?

src/components/ProfileContentActivity.js
```diff
@@ -0,0 +1,77 @@
...
+  };
+});
+
+const ProfileContentActivity = ({ address }) => {
+  const classes = useStyles();
+  const dispatch = useDispatch();
+  const mutualActivities = useSelector(
+    (state) => state.activity.mutualActivities.activities,
+  );
+  const isMoreAvailable = useSelector(
```
louilinn, line 26: This parameter does not seem to work accurately as there is a load more button even when all are fetched (e.g. 1 or 4 / less than 10 items). I think the regular activity log might also have this problem. But still, consider adding logic to turn load more off if first fetch <10. I think product might ask us to do this later otherwise

src/store/activity/reducers.js
```diff
@@ -176,6 +188,48 @@ const activityReducer = (state = initialState, action) => {
...
+    case ActionTypes.ACTIVITIES_MUTUAL_LOAD_MORE:
+      return update(state, {
+        mutualActivities: {
+          isError: { $set: false },
+          isLoadingMore: { $set: true },
+        },
+      });
+    case ActionTypes.ACTIVITIES_MUTUAL_LOAD_MORE_SUCCESS: {
+      // Nothing more to add ..
+      if (action.meta.activities.length === 0) {
```
louilinn: try replacing the condition with
 `action.meta.activities.length < action.meta.offset` 
 or similar to have the load more button disappear when less than than 10 are fetched
 Maybe this is not offset but.. page size? This constant should exist somewhere for the call to the core, use the same constant

src/utils/debug.js
```diff
@@ -52,11 +52,11 @@ export function translateErrorForUser(error) {
...
     if (error.code == ErrorCodes.INSUFFICIENT_FUNDS) {
       text = translate('ErrorCodes.CoreErrorInsufficientFunds');
-    }
-
-    if (error.code == ErrorCodes.SAFE_NOT_FOUND) {
+    } else if (error.code == ErrorCodes.SAFE_NOT_FOUND) {
       //  error: `Could not find Safe with address ${safeAddress}`,
       text = error.message;
+    } else {
+      text = `${error.message}. ${translate('ErrorCodes.GeneralErrorMessage')}`;
```
louilinn, line 59: Why? See line 62?

src/store/activity/actions.js
```diff
@@ -276,3 +350,9 @@ export function resetActivities({ isClearingStorage = true } = {}) {
     type: ActionTypes.ACTIVITIES_RESET,
   };
 }
+
+export function resetMutualActivities() {
```
louilinn: Where is this used? and why is it needed?

louilinn (changes requested):
Great job Miko :)
I have some concerns about load more and some other comments

src/components/ActivityStream.js
mikozet: corrected

src/components/ActivityStream.js
mikozet: At the beginning I didn't share exact data structure like we have in trust and transaction. Later it changed but I decided to leave it as News will have different data structure and we need to be more flexible here

src/components/ButtonAction.js
mikozet: yes, I checked and it worked fine :)

src/components/ProfileContentActivity.js
mikozet: Hi, yes. I didn't touch that as it is standard behaviour at the moment and it requires some testing etc so not that super quick fix. We may fix it after there is a task for that.

src/store/activity/reducers.js
mikozet: yeah I have seen it but decided to leave it as it is sot it is consistence with activities.
But since you commented I decided to than repair it over there as well.

So I add comment to this task to change it's scope:
https://github.com/CirclesUBI/circles-myxogastria/issues/272

It is not solved now 100% but it improved a lot..

src/utils/debug.js
mikozet: There are more CoreError and we do not have case for every type.
I added one more case when there are invalid options for example here it will populate:
"safeAddress" is missing in options. Please try with different values or refresh page or try again later"

src/store/activity/actions.js
mikozet: corrected

mikozet:
Added two checkboxes in the description of task

src/store/activity/actions.js
```diff
@@ -22,7 +22,7 @@ import { formatCirclesValue } from '~/utils/format';
 
 const { ActivityTypes, ActivityFilterTypes } = core.activity;
 
-export const PAGE_SIZE = 10;
+export const PAGE_SIZE = 2;
```
louilinn: Was this changed for testing purposes or is this an intentional change?

src/utils/debug.js
louilinn: But logically it will be caught in the last else either way no?

src/store/activity/actions.js
louilinn: is this same constant also used in the request to the core? That is important

louilinn (changes requested):
Follow up comments :)

src/store/activity/actions.js
mikozet: It should be 10, I corrected it. `PAGE_SIZE` is send to core

src/utils/debug.js
mikozet: Yes, but I am adding those types so we are more aware of them and have more control over what do we want to display for the user.

juanenrisley: I understand what @mikozet is trying to do here. @louilinn if the error is instance of CoreError, it will never enter the last `else` statement thus will be not logged.

What we could do to not dirt the code with so much if/else is just to initialize the variable `text` with a default value like this ```let text = `${error.message}. ${translate('ErrorCodes.GeneralErrorMessage')}`;``` and then delete both `else` statements with same declaration.

louilinn:
Feedback from Lucia: The component of profile content should also be visible on profiles of shared wallets.
- [x] figure out why the mutual connections and shared activity doesnt show up for SWs
- [x] fix

src/utils/debug.js
louilinn: Yes ok I see now x) 
Yeah that would be the better option but it's ok to leave also. But this is confusing code :p

src/store/activity/actions.js
```diff
@@ -22,7 +22,7 @@ import { formatCirclesValue } from '~/utils/format';
 
 const { ActivityTypes, ActivityFilterTypes } = core.activity;
 
-export const PAGE_SIZE = 2;
+export const PAGE_SIZE = 10;
```
louilinn, line 25: Can you please make sure that this is the same parameter as is sent to the core request, so that if changed, the pagination to the request is also changed?
Otherwise if we change this later on, some notifications will not be shown or similar, if I'm not mistaken. I haven't thought it through completely, but it can mess up behaviour if the next person editing thinks this changes page size in the request. It is important to use propagate such a variable to the request, otherwise it is deceiving

louilinn (dismissed):
This plus the comments above from Lucia

src/utils/debug.js
mikozet: I liked @juanenrisley idea I corrected it

mikozet:
> Feedback from Lucia: The component of profile content should also be visible on profiles of shared wallets.
> 
> * [x]  figure out why the mutual connections and shared activity doesnt show up for SWs
> * [x]  fix

It was not showing as we were checking token deployment status. For Organisation accounts it is ZERO_ADDRESS and in that case we didn't show it. I added a confdition regarding organisation account

src/store/activity/actions.js
mikozet: as I wrote before it is - it is in line 311. Code below:

```
const { activities, lastTimestamp } = await core.activity.getLatest(
        safe.currentAccount,
        ActivityFilterTypes.DISABLED,
        PAGE_SIZE,
        activity.lastTimestamp,
        offset,
        otherSafeAddress,
      );

juanenrisley (approved):
👍🏿 LGTM

JacqueGM (approved):

juanenrisley (approved):
👍🏿

PR #748, opened 2023-10-06
Make mutual connections accurate for all accounts

louilinn:
- [x] update core version
- [x] add mutual connections count to all mini-profiles
- [x] call checkTrustStatus in Tokens view (Balance Breakdown) to see correct numbers in mini-profiles there


juanenrisley (approved):
👍🏿

PR #752, opened 2023-10-10
Small fixes

### LINKED ISSUE #181: Should we add also lower case characters ([a-z]) in "formUsernameInvalidFormat"?
In the text for the views, the [field "formUsernameInvalidFormat"](https://github.com/CirclesUBI/circles-myxogastria/blob/organizational-wallets/locales/en.json#L326) might suggest that lower case characters are not valid.

louilinn:
- text edits
- gradient fix mistake

Closes: https://github.com/CirclesUBI/circles-myxogastria/issues/181


juanenrisley (approved):
👍🏿

PR #794, opened 2024-09-09
Additional German translation

lotteam006:
Additional German translation


mikozet (approved):

mikozet:
Since we found a lot of mistakes in translation in Spanish version.. we would like first some native speaker or someone with a good sense of this language knowledge to review it

